PR #2, opened 2020-04-13
secp256k1: initial build

benjamincburns:
Initial native build for secp256k1, including circleci config

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+#############################
+######### Variables #########
+#############################
+
+# Edit this variable to change the build options for secp256k1
+SECP256K1_BUILD_OPTS="--enable-module-recovery"
+
+#############################
+####### End Variables #######
+#############################
+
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -0,0 +1,66 @@
...
+cd "$SCRIPTDIR/secp256k1/bitcoin-core-secp256k1"
+
+# delete old build dir, if exists
+rm -rf "$SCRIPTDIR/secp256k1/built" || true
+
+
+
+make clean && \
+  ./autogen.sh && \
+  ./configure --prefix="$SCRIPTDIR/secp256k1/built" $SECP256K1_BUILD_OPTS && \
```
shemnon, line 57: Why `built`?  The traditional destinations I've seen are `build` (gradle, ant) and `target` (maven, rust)

benjamincburns: happy to change it if you want - might be best if you set it to whatever is most convenient when you add the gradle build?

Also I realized just now that I likely should be setting the prefix based on OS/Arch combo, especially because of the eventual arm build...

PR #6, opened 2020-04-15
Multiplatform build with artifacts publishing to bintray

benjamincburns:
Includes changes to the CircleCI configs to support building native dependencies across Linux (x86_64), and macOS. Will update in a future PR with support for Linux (ARM32) and Windows.

Signed-off-by: Ben Burns <803016+benjamincburns@users.noreply.github.com>


benjamincburns:
Note that I've tested the CircleCI config and publish extensively. I'm not sure if this build is publicly visible, but here's an example of it passing and publishing to the test bintray repo that we use at PegaSys: https://circleci.com/workflow-run/87807689-55c3-478d-968c-dbe714b3bbc2

PR #13, opened 2020-04-23
Add BLS12-381 library

shemnon:
Add the Matter Labs EIP2537 library to handle BLS12-381 precompiles.

### DIFF
--- a/bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
+++ b/bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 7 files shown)


bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
```diff
@@ -0,0 +1,58 @@
...
+
+  static {
+    boolean enabled;
+    try {
+      Native.register(LibEthPairings.class, "eth_pairings");
+      enabled = true;
+    } catch (final Throwable t) {
+      enabled = false;
+    }
+    ENABLED = enabled;
```
RatanRSur, line 33: why do we need this intermediate boolean `enabled`? something weird about `static` blocks?

shemnon: the field is final, and the compiler analysis considers catch blocks as a possible second setting because it doesn't calculate where in the try block the exception may be thrown.

PR #21, opened 2020-09-29
Add EIP196/197 methods to matterlabs library

shemnon:
Ad EIP 196 (BN add/mul) and EIP197 (BN pairing).  Performance is notably
better than the sputnikvm code.

### DIFF
--- a/bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
+++ b/bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
@@ -20,6 +20,7 @@ import com.sun.jna.Native;
 import com.sun.jna.ptr.IntByReference;
 
 public class LibEthPairings implements Library {
+  @SuppressWarnings("WeakerAccess")
   public static final boolean ENABLED;
 
   static {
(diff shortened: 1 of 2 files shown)


bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
```diff
@@ -55,4 +56,21 @@ public static native int eip2537_perform_operation(
...
+
+  public static final int EIP196_PREALLOCATE_FOR_ERROR_BYTES = 256;
+
+  public static final int EIP196_PREALLOCATE_FOR_RESULT_BYTES = 64;
+
+  public static final byte EIP196_ADD_OPERATION_RAW_VALUE = 1;
+  public static final byte EIP196_MUL_OPERATION_RAW_VALUE = 2;
+  public static final byte EIP196_PAIR_OPERATION_RAW_VALUE = 3;
+
+  public static native int eip196_perform_operation(
```
abdelhamidbakhta, line 68: i would say that the names of the parameters are not very explicit. i guess you took those from the native implementation ?

gradle.properties
```diff
@@ -1 +1 @@
-version=0.2.1-SNAPSHOT
+version=0.3.0-SNAPSHOT
```
abdelhamidbakhta, line 1: is there a versioning policy ? do we change the second digit of the version each time there is a new native function ?

abdelhamidbakhta (approved):
LGTM. Left some minor comments. I have only look the Java changes and not the submodule changes

gradle.properties
shemnon: SemVer.  New Long-term APIs should be minor.  https://semver.org/

bls12-381/src/main/java/org/hyperledger/besu/nativelib/bls12_381/LibEthPairings.java
shemnon: Yes, this name must match the library.

PR #30, opened 2021-02-06
Revise Artifactory Config

shemnon:
Use the correct artifactory context url


macfarla (approved):
LGTM

PR #51, opened 2021-11-15
use non-blocking PRNG to generate secp256k1 context

taccatisid:
This uses `java.util.Random.nextBytes` instead of `java.security.SecureRandom.generateSeed` to generate the seed used to "randomize" the secp256k1 context. The later might block indefinitely on some platforms when the OS' RNG runs out of entropy.

`java.util.Random` is a much weaker source of randomness (2^48 cycle, much less under real world conditions due to imperfect seed). It is however sufficient in this case: The context randomization is only used as an additional precaution against side channel attacks in case the compiler ruins the constant time property of the underlying code.

Signed-off-by: Taccat Isid <taccatisid@protonmail.com>

### DIFF
--- a/secp256k1/src/main/java/org/hyperledger/besu/nativelib/secp256k1/LibSecp256k1.java
+++ b/secp256k1/src/main/java/org/hyperledger/besu/nativelib/secp256k1/LibSecp256k1.java
@@ -16,7 +16,7 @@
 package org.hyperledger.besu.nativelib.secp256k1;
 
 import java.nio.ByteBuffer;
-import java.security.SecureRandom;
+import java.util.Random;
 
 import com.sun.jna.Callback;
 import com.sun.jna.Library;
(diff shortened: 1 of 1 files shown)


taccatisid:
Fixes https://github.com/hyperledger/besu/issues/3013.

lucassaldanha (approved):

taccatisid:
Regarding the question, **whether it is secure to reduce the strength of the RNG** here.

Looking at the question from a security risk perspective: To make a high-entropy randomness source (_~256b_) an effective protection, but a low entropy-source (_32b-48b_) not, all of the following things have to be true:
1. the compiler finds a way to turn constant-time code into non-constant time code
1. the non-constant time happens in the curve point arithmetic itself, not in the outside use of key or ciphertext material
2. the side channel attack is not susceptible to an analog measurement noise floor given a signal to noise ratio < 1:2⁻³²
3. the side channel attack has a brute force component that is cheap enough to be economical after amplification by a factor 2³²
I would consider each statement unlikely to extreme unlikely and can go into detail why if you want. Multiplying those probabilities leads to near impossibility.

This calculus would change a little if `java.util.Random` would deliver an extremely small amount of entropy--like the Debian OpenSSL debacle where entropy was only _16b_. But even then we are most likely in safe waters due to 1-3 still holding.

PR #56, opened 2022-03-20
Add new ipa-multipoint library integration

atoulme:


gballet:
Context for this PR: verkle trees replace hashes with polynomial commitments. These polynomials are computed over an elliptic curve, called bandersnatch. This PR provides the function that takes the coefficients of the polynomial and evaluates it: It's, therefore, the basic building block of verkle trees.

ipa-multipoint/ipa_multipoint_jni/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+repository = "https://github.com/hyperledger/besu-native"
+edition = "2018"
+
+[dependencies]
+ipa-multipoint = { path = "../ipa_multipoint" }
+bandersnatch = "0.1.1"
+jni = { version = "0.19.0", features = ["invocation"] } # We use invocation in tests.
+rand_chacha = "0.3.1"
+rand_core = "0.6.3"
+rand = "0.8.5"
```
gballet, line 18: It seems that these libraries are no longer needed - better remove them as downloading them will slow the build down.

ipa-multipoint/ipa_multipoint_jni/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+name = "ipa_multipoint_jni"
+version = "0.1.0"
+description = "IPA Multipoint Crypto Library"
+license = "Apache-2.0"
+authors = ["Antoine Toulme <antoine@lunar-ocean.com>"]
+repository = "https://github.com/hyperledger/besu-native"
+edition = "2018"
+
+[dependencies]
+ipa-multipoint = { path = "../ipa_multipoint" }
```
gballet, line 13: I recommend using the `git =` key so that it will go get the latest version. This repo is mostly stable, but some tweaks are still in the pipeline.

```suggestion
ipa-multipoint = { git = "https://github.com/crate-crypto/ipa_multipoint", branch = "develop" }
```

gballet (commented):
Added a couple nitpicks

ipa-multipoint/ipa_multipoint_jni/Cargo.toml
atoulme: fixed

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "secp256r1/besu-native-ec"]
 	path = secp256r1/besu-native-ec
 	url = git@github.com:ConsenSys/besu-native-ec.git
+[submodule "ipa-multipoint/ipa_multipoint"]
```
gballet, line 10: you no longer need `ipa_multipoint` as a submodule, as it's referenced by `ipa_multipoint_jni` in its `Cargo.toml`.

garyschulte (changes requested):
I can't meaningfully comment on the rust impl.  Feedback about multi-arch darwin jars should be easy to implement

build.sh
```diff
@@ -96,6 +96,26 @@ EOF
...
+
+  cd "$SCRIPTDIR/ipa-multipoint/ipa_multipoint_jni"
+
+  # delete old build dir, if exists
+  rm -rf "$SCRIPTDIR/ipa-multipoint/build" || true
+  mkdir -p "$SCRIPTDIR/ipa-multipoint/build/lib"
+
+  cargo clean
+
+  cargo build --lib --release
```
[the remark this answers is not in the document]
atoulme: OK I'll take a look.

ipa-multipoint/src/main/java/org/hyperledger/besu/nativelib/ipamultipoint/LibIpaMultipoint.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright The Machine Consultancy LLC
```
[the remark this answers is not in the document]
atoulme: OK.

build.sh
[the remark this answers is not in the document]
atoulme: ok should be good I guess?

garyschulte (approved):
LGTM

ipa-multipoint/src/main/java/org/hyperledger/besu/nativelib/ipamultipoint/LibIpaMultipoint.java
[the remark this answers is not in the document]
shemnon: This is what the [Hyperledger Copyright and License Policy](https://wiki.hyperledger.org/display/TSC/Copyright+and+License+Policy) calls for.

ipa-multipoint/src/main/java/org/hyperledger/besu/nativelib/ipamultipoint/LibIpaMultipoint.java
```diff
@@ -0,0 +1,26 @@
...
+package org.hyperledger.besu.nativelib.ipamultipoint;
+
+/**
+ * Java interface to ipa-multipoint.
+ *
+ * Allows to compute commitments for Verkle tree nodes.
+ */
+public class LibIpaMultipoint {
+
+  public static native byte[] commit(byte[] input1, byte[] input2, byte[] input3, byte[] input4);
```
shemnon, line 25: Are we always only committing 4 values?  Do each of the values have meaning? I think a bit more javadoc would be appropriate if it is going into a general release.

atoulme: 1. yes, always 4 values.
2. They form a lagrange basis point - and that's all I know. @gballet can you help document?

I'm not aware of a draft spec for ipa multipoint, I just built this on top of the rust lib as you can see, and it's not documenting this either. I took the route of admitting that I had no idea what I was looking at (I'm really not a mathematician) and that I'd make it work, but sure, it'd be nice to know more about the ins and outs.

shemnon: This is a crypto library.  We should either be able to explain it or point to papers/libraries/written material that explain it.  How can we even be sure we are using the crypto primitive correctly?

atoulme: Pretty much, I took the test vector in rust and made it work for Java IIRC.

gballet: Some clarifications here:

1. the curve is called bandersnatch and is described here https://eprint.iacr.org/2021/1152.pdf
2. this `commit` function is simply the evaluation of a polynomial at a specific point. This is what polynomial commitments mean: the evaluation of a polynomial at a given point, that you can later use to prove that you know the full polynomial.
3. In the context of verkle trees, there are two such polynomial evaluations that can occur:
  * When calculating a so-called "pedersen hash", in which a degree 3 polynomial is evaluated
  * When calculating the commitment to a tree node, in which a degree 255 polynomial is evaluated

This first function therefore corresponds to the first case: the evaluation of a polynomial of degree 3 (uniquely defined by 4 values). For your documentation effort, maybe the function should be renamed to something like `commitToDegree3Polynomial` and the 4 values be renamed `y0`, `y1`, `y2` and `y3` as they are the y values that this polynomial takes at x=0,1,2,3.

Alternatively, we could imagine modifying these values to take a polynomial of any degree, but I don't think that it's possible for JNI to pass `[][]byte` values, is it?

atoulme: We could change the Java signature, but for now let's address the problem of the missing documentation. I am going to take your comments and turn them into code changes, such as Java comments, to help complete the Java method signature and get us past this current hurdle.

PR #70, opened 2022-08-03
fix the dependency to the lib to a specific commit, d5590ef5

atoulme:


garyschulte (approved):
👍

PR #71, opened 2022-08-03
Add github build to replace circleCI build

atoulme:


shemnon (dismissed):
Missing linux arm build - https://github.com/hyperledger/besu-native/pull/71/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L127

.github/workflows/build.yml
```diff
@@ -0,0 +1,245 @@
...
+      - uses: actions/upload-artifact@v3.1.0
+        with:
+          name: secp256r1 native build artifacts
+          path: secp256r1/besu-native-ec/release/
+      - uses: actions/upload-artifact@v3.1.0
+        with:
+          name: ipa-multipoint native build artifacts
+          path: ipa-multipoint/build/
+  native-build-linux-arm64:
+    runs-on: besu
```
shemnon, line 51: Can we get a different name?  This to me implies a server with besu on it, no a build environment. is it ok if it's named `besu-ci` like the github actions config?

shemnon:
LGTM once the final build error is fixed.  Since there are major blake2f changes I think @fab-10 should provide the final review.

atoulme:
The final build error will not be solved - the CircleCI build will be failing until we unplug the repository.

.github/workflows/build.yml
atoulme: You might be able to change in github in the runner tags. I cannot access the admin interface.

build.sh
```diff
@@ -58,7 +61,11 @@ build_blake2bf() {
 EOF
 
   if [[ "$OSTYPE" == "linux-gnu" ]];  then
-    cd "$SCRIPTDIR/blake2bf/$( arch )"
+    if [[ $(uname -m) == 'aarch64' ]]; then
```
fab-10, line 64: I am a bit confused by this:

- if `$(uname -m) == 'aarch64'` then `arm64` dir
- if `$(uname -m) == 'arm64'` then `aarch64` dir

it seems clearer to simply swap the dir names, or am I missing something?

build.sh
```diff
@@ -71,6 +78,19 @@ EOF
     mkdir -p "$SCRIPTDIR/blake2bf/build/${OSARCH}/lib"
     mv libblake2bf.so "$SCRIPTDIR/blake2bf/build/${OSARCH}/lib"
   fi
+  if [[ $(uname -m) == 'arm64' ]]; then
```
fab-10, line 81: my understanding is this adds support for building on ARM based MAC, right? 
it seems that this part could be easily merged with the linux part, extending the if to include `"$OSTYPE" == "darwin"*` 

Moreover if it possible to swap `arm64` dir with `aarch64` dir, there should be no need to check the architecture

fab-10 (commented):
Thanks for adding support for building Blake2bf on Darwin.
I have just a couple of comments on optimizing the build script, if possible, otherwise it is fine for me.

build.sh
shemnon: It doesn't create a dylib, so it's not a complete mac build.

atoulme: This adds support for arm build on mac m1, which is the besu runner. I wasn't sure linux arm64 was compatible with apple aarch64.

build.sh
atoulme: this is to override - on m1 besu self hosted runner, I found the `$(arch)` call returned an invalid value. Instead, I catch early that if `uname -m` is aarch64, then we need to use the arm64 sources.

build.sh
fab-10: I do not have access to a Mac M1 to verify if the script could be generalized in the way I proposed in this PR https://github.com/atoulme/besu-native/pull/3, it will be great if you can try it

fab-10 (approved):
LGTM, just suggesting a generalization of the Black2bf build script here https://github.com/atoulme/besu-native/pull/3, in case it works also on Mac M1, but it is non blocking.

PR #73, opened 2022-08-15
No build artifacts

shemnon:
There should be no build artifacts checked into source control.
* Remove one that was accidentally checked in
* Add .gitignore settings to cover common build artifacts


macfarla (approved):
LGTM

PR #79, opened 2022-09-05
Ship both x86 and aarch64 darwin libs

atoulme:

### DIFF
--- a/ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/LibIpaMultipointTest.java
+++ b/ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/LibIpaMultipointTest.java
@@ -17,19 +17,14 @@ package org.hyperledger.besu.nativelib.ipa_multipoint;
 
 import org.apache.tuweni.bytes.Bytes;
 import org.hyperledger.besu.nativelib.ipamultipoint.LibIpaMultipoint;
-import org.junit.BeforeClass;
 import org.junit.Test;
-
-import java.io.File;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 public class LibIpaMultipointTest {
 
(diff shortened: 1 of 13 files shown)


ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/LibIpaMultipointTest.java
```diff
@@ -17,36 +17,14 @@
...
-    } else if (macFile.exists()) {
-      System.load(macFile.getAbsolutePath());
-    } else {
-      throw new RuntimeException("could not setup jni path for test");
-    }
-  }
-
   @Test
-  public void testCallLibrary() {
+  public void testCallLibrary() throws Exception {
```
shemnon, line 27: ?? Why add the generic `throws Exception` ?? None of the called apis were changed.

shemnon (approved):
Looks good for the most part.

Except that the Blake2F lib performs worse than the Java lib right now on mac. Performance is out of scope for this PR.

garyschulte (commented):
Are we not able to build a lipo/multi-arch darwin lib for blake2f in the same way we do for the other besu-native libs?

ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/LibIpaMultipointTest.java
atoulme: Probably an artifact from the back and forth during testing. I can remove that change, sure.

atoulme:
jna now uses separate folders for darwin, so you can't use a multi-arch lib.
And blake2bf has some ever so slightly differences in imports between m1 and arm, and required a change in the code for imports. You can probably reconcile and use a precompile condition like `ifdef __APPLE__`. I picked that up a bit late.

atoulme:
I have verified locally that the m1 lib for ipa-multipoint now loads properly with the zulu java 17 SDK.

PR #82, opened 2022-09-21
Disable arm blake2b

shemnon:
Benchmarking shows the blake on arm code is slower than the java versions running on arm, so don't build it.


garyschulte (approved):
expedient way to omit blake2f on arm.  

This might be worth a TODO and an issue to address it in the future so we do not lose track.

shemnon:
https://github.com/hyperledger/besu-native/issues/83

PR #85, opened 2022-09-21
Restore Mac ECDSA libs

shemnon:
Restore the generation of the two ECDSA libs in the new darwin-<arch> format.


shemnon:
Fixes error in #79 where these files were not migraged in the build file.

garyschulte (approved):

PR #98, opened 2022-12-02
Add module for low level arithmetic

shemnon:
Add a module for low level arithmetic, for performance. First two operations are modexp precompile and mul operation.


gradle.properties
```diff
@@ -1 +1 @@
-version=0.6.2
+version=0.7.0-SNAPSHOT
```
atoulme, line 1: That probably needs its own PR

shemnon: should have been part of 0.6.2 release.

shemnon: https://github.com/hyperledger/besu-native/pull/99

macfarla (approved):

PR #107, opened 2023-05-02
rename mimc to gnark

matkt:
rename mimc library to gnark


siladu (approved):
Not familiar with the code, but appears to be a reasonable renaming operation

macfarla (approved):

PR #118, opened 2023-09-13
Add k1 normalize signatures method

jasperpotts:
The normalize signatures method was missing in Java wrapper. It is needed so verify signatures can be used will all signatures. Without it only a subset of signatures are valid.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,1 @@
-version=0.8.2-SNAPSHOT
+version=0.8.3-SNAPSHOT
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -1 +1 @@
-version=0.8.2-SNAPSHOT
```
shemnon, line 1: This shouldn't be bumped.  0.8.2 has yet to be released.

jasperpotts: Restored version

shemnon (approved):

PR #125, opened 2023-11-03
FIX: javadoc problems, removes <-> chars

thomas-quadratic:
Javadoc was failing because of HTML-incompatible characters.
They were removed


matkt (approved):
LGTM

PR #127, opened 2023-11-24
add update commitment to ipa_multipoint JNI

dragan2234:
Changing out ipa_multipoint API to these functions:

- pedersenHash: same as before (not changed)
- commit: expects up to 256 BE 32byte Fr values, returns Commitment serialized to Fp.
- updateCommitment: expects Commitment, diff value, and index, returns Commitment serialized to Fp
- groupToField: expects Commitment serialized to Fp, returns Fr

Everything expects BE bytes, and returns BE bytes.

I still think Java is working on BE bytes by default. Please re-run `TestPedersenCommitment` and put a breakpoint. Example and screenshot:

995CEEBE-40F4-4DAD-B511-8706FA202285

so [28, -127, 74 ....

is 1c,81,4a in hex

Only thing that is working on LE is "serialize" call from arkworks


dragan2234:
This is ready for review @matkt @thomas-quadratic .

Note that it will break some things in besu-verkle-trie repo since we changed the API a bit per the description above.

So in order to get "commitment" what is described as commitment in besu-verkle-trie you just call `LibIpaMultipoint.commit()` and you get the C serialized to bytes(Fp).

Now if you want to reuse in parent or save it as "a hash" you just call `LibIpaMultipoint.groupToField()`.

Same thing applies for `LibIpaMultipoint.updateCommitment()`.

thomas-quadratic:
This PR addresses issues #129 #128

PR #131, opened 2023-12-18
Linking error potential fix

neotheprogramist:
## Description
This pull request addresses a critical linking error encountered in the Hyperledger Besu native library. The issue came to light following a modification in the method invocation within the `PedersenHasher` class, where `LibIpaMultipoint.commit` was changed to `LibIpaMultipoint.commit_root`. As a result of this change, a `java.lang.UnsatisfiedLinkError` is triggered when calling the `commit_root` method, with the error message: `'byte[] org.hyperledger.besu.nativelib.ipamultipoint.LibIpaMultipoint.commit_root(byte[])'`.

### Proposed Fix
The fix proposed in this pull request involves renaming the `commit_root` method from snake_case to camelCase. This renaming aims to:

- Align with Java's standard naming conventions.
- Potentially resolve the linking issues that are currently being faced.

### Context
While other methods in the `LibIpaMultipoint` class, such as `commit` and `pedersenHash`, continue to operate as expected, it is believed that the naming convention employed for `commit_root` may be contributing to the linking error.


matkt:
seems to work . could we add a small description in the PR ?

thomas-quadratic:
Shouldn't we have a test in ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/ for this method as well ?

thomas-quadratic:
Good with me. Thanks!

ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/CommitRootTest.java
```diff
@@ -0,0 +1,46 @@
+package org.hyperledger.besu.nativelib.ipa_multipoint;
+
```
matkt, line 2: Seems good, could you just add a header like this file https://github.com/hyperledger/besu-native/blob/main/gnark/src/main/java/org/hyperledger/besu/nativelib/gnark/LibGnark.java#L1

matkt (approved):
Lgtm

thomas-quadratic:
Good for me as well, thanks @neotheprogramist

PR #133, opened 2024-01-04
WIP: Expose verkle prove call

dragan2234:
WIP exposing prove call.

https://github.com/hyperledger/besu-native/pull/133/files#diff-6c182f542873a0b4e1718eec91be7decc4f4ff80602e2da0c11b9e68e93c55beR196

I would rather first wait for https://github.com/hyperledger/besu-native/pull/127 and test the trie with this crypto API and computing a root commitment before merging the proving/verifying calls, but as requested here's the draft. cc @thomas-quadratic @matkt


dragan2234:
Closing as it's stale since we moved to a bit different API

PR #138, opened 2024-01-19
feat: optimize multipoint api

dragan2234:
This is ready for testing. I'd still make more changes, but that needs to be in sync with besu-verkle-trie repo, so for now here it is, together with https://github.com/hyperledger/besu-verkle-trie/pull/31

cc @matkt @thomas-quadratic

### DIFF
--- a/ipa-multipoint/ipa_multipoint_jni/Cargo.toml
+++ b/ipa-multipoint/ipa_multipoint_jni/Cargo.toml
@@ -17,20 +17,15 @@ name = "ipa_multipoint_jni"
 version = "0.1.0"
 description = "IPA Multipoint Crypto Library"
 license = "Apache-2.0"
-authors = ["Antoine Toulme <antoine@lunar-ocean.com>"]
+authors = ["Antoine Toulme <antoine@lunar-ocean.com>", "Dragan Pilipovic https://github.com/dragan2234", "thomas-quadratic https://github.com/thomas-quadratic"]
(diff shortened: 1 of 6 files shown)


thomas-quadratic:
The CI/CD build seems to be failing.
On my computer, it fails with the following error:

> Task :ipa-multipoint:compileTestJava FAILED
/besu-native/ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/LibIpaMultipointTest.java:79: error: method pedersenHash in class LibIpaMultipoint cannot be applied to given types;
        Bytes result = Bytes.of(LibIpaMultipoint.pedersenHash(total));
                                                ^
  required: byte[],long
  found:    byte[]
  reason: actual and formal argument lists differ in length
1 error

Could you please investigate this ? I think that there has been a change in the class but not in the test class.

dragan2234:
@thomas-quadratic This is an old PR, main purpose is the usage of the pointer https://github.com/hyperledger/besu-native/pull/138/files#diff-6c182f542873a0b4e1718eec91be7decc4f4ff80602e2da0c11b9e68e93c55beR35 and https://github.com/hyperledger/besu-native/pull/138/files#diff-6c182f542873a0b4e1718eec91be7decc4f4ff80602e2da0c11b9e68e93c55beR57 which should be built on top of what we have right now.

But maybe we'll go away from this design, still not sure, but I think it's useful to have this code somewhere for now

PR #139, opened 2024-01-19
chore: Switch to `ffi_interface`

kevaundray:
This switches to the new `ffi_interface` folder and avoids any interface changes by using a global static.

The global static should be replaced with a pointer as is done in #138


kevaundray:
Depends on #142

.github/workflows/build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           export PATH=$GOROOT/bin:$PATH
           # rust dependencies
           export CARGO_HOME="$HOME/.cargo"
-          curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.68.2
+          curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.75.0
```
kevaundray, line 30: This change was added because CI uses stable and matt was getting an unstable error from one of the newer libraries for a feature which should be stable as of 1.75. (1.68 is quite old)

ipa-multipoint/src/test/java/org/hyperledger/besu/nativelib/ipa_multipoint/LibIpaMultipointTest.java
```diff
@@ -55,24 +55,24 @@ public void testCallLibraryWithManyElements() {
 
     @Test
     public void testCallLibraryWithMaxElements() {
-        Bytes32 element = Bytes32.fromHexString("0xd36f20567f74f607d9252186ff8efed04de4578d1ddb3de4fe6c5e4249e0045b");
+        Bytes32 element = Bytes32.fromHexString("0x006f20567f74f607d9252186ff8efed04de4578d1ddb3de4fe6c5e4249e0045b");
         Bytes32[] arr = new Bytes32[256];
         for (int i = 0; i < 256; i++) {
             arr[i] = element;
         }
```
kevaundray, line 62: These changes are from #142

matkt (approved):
LGTM

PR #141, opened 2024-01-24
fix: fix the commit for verkle dependency

kevaundray:
This fixes the commit for verkle-trie, since pointing to master can make compiles non-deterministic.

This will change soon to using ffi_interface, though while that PR is not merged this is a fix.


matkt (approved):
LGTM

PR #143, opened 2024-01-28
chore: update to a version of rust-verkle which uses little endian

kevaundray:
This uses a version of rust-verkle which uses little-endian.

This is not a breaking change because we change the endianness in the jni library. We can make a separate PR to change besu-verkle and other components to use little endian. Doing it this way reduces the amount of moving parts to making this change.


kevaundray:
This is based off #139 -- once that is merged the diff will be smaller

matkt (approved):
LGTM

PR #144, opened 2024-01-28
chore: Remove direct dependency on ipa-multipoint

kevaundray:
This removes the direct dependency on ipa-multipoint, so that only one import is needed; ffi-interface.


kevaundray:
Depends on #143

ipa-multipoint/ipa_multipoint_jni/src/lib.rs
```diff
@@ -45,130 +39,92 @@ pub extern "system" fn Java_org_hyperledger_besu_nativelib_ipamultipoint_LibIpaM
...
-    scalar_bytes.reverse();
-
-    return env.byte_array_from_slice(&scalar_bytes).expect("Couldn't convert to byte array");
-}
+pub extern "system" fn Java_org_hyperledger_besu_nativelib_ipamultipoint_LibIpaMultipoint_commit(
+    env: JNIEnv,
+    _class: JClass<'_>,
+    input: jbyteArray,
+) -> jbyteArray {
+    let mut input = env
```
thomas-quadratic, line 81: Should we use the same way to read inputs as line 46 ? It is safer I believe
let mut input: [u8; 64] = match input.try_into() {
        Ok(input) => input,
        Err(_) => {
            env.throw_new(
                "java/lang/IllegalArgumentException",
                "Invalid input length. Should be 64-bytes.",
            )
            .expect("Failed to throw exception");
            return std::ptr::null_mut(); // Return null pointer to indicate an error
        }
    };

ipa-multipoint/ipa_multipoint_jni/src/lib.rs
```diff
@@ -45,130 +39,92 @@ pub extern "system" fn Java_org_hyperledger_besu_nativelib_ipamultipoint_LibIpaM
...
-    
-    // Committing all values at once.
-    let bases = CRS::new(n_scalars, PEDERSEN_SEED);
-    let commit = multi_scalar_mul(&bases.G, &scalars);
+pub extern "system" fn Java_org_hyperledger_besu_nativelib_ipamultipoint_LibIpaMultipoint_commitRoot(
+    env: JNIEnv,
+    _class: JClass<'_>,
+    input: jbyteArray,
+) -> jbyteArray {
+    let mut input = env
```
thomas-quadratic, line 107: Same comment, we should have a uniform way of converting inputs and outputs from/to Java.

ipa-multipoint/ipa_multipoint_jni/src/lib.rs
```diff
@@ -45,130 +39,92 @@ pub extern "system" fn Java_org_hyperledger_besu_nativelib_ipamultipoint_LibIpaM
...
 
-    // Serializing using first affine coordinate
-    let commit_bytes = commit.to_bytes();
+    let committer = &CONFIG.committer;
+    reverse_scalars_endian(&mut input);
 
-    return env.byte_array_from_slice(&commit_bytes).expect("Couldn't convert to byte array");
-}
+    let commitment = ffi_interface::commit_to_scalars(committer, &input).unwrap();
+    let hash = ffi_interface::deprecated_serialize_commitment(commitment);
```
thomas-quadratic, line 115: Is deprecated_serialize_commitment deprecated ? 
I suppose this is the compressed serialisation.

thomas-quadratic (commented):
Hi thanks for all the improvements. This is looking good now.

I believe this PR breaks the Java-side because it assumes that it passes as input uncompressed serialised commitments. If so, this is ok. However, while we are at it, I would also make the assumption that inputs and outputs from/to Java are little endian.

matkt:
tried this version and I have a valid state root for genesis block 637 ms with this current implementation to generate the root @kevaundray @thomas-quadratic @dragan2234

thomas-quadratic:
Ok, so once the DCO check passes, I think we can merge.
If there are other things that come up (commitment format, endianness), we can create new PRs for them.

ipa-multipoint/ipa_multipoint_jni/src/lib.rs
kevaundray: There are currently two ways to serialize a group element -- I'm looking to deprecate this one on the API level -- it is used in other places, but ideally not directly by the caller as it is being used now

ipa-multipoint/ipa_multipoint_jni/src/lib.rs
thomas-quadratic: Ah, I missed that!
Nonetheless, it is strange that in one case we throw a Java Error and in the other we don't.
However, I am not sure how error propagation works on the JNI side.
I think that we can do as is and if the need emerge to handle errors in another way, we will do it then.

matkt: I will prefer to  use 
```
env.throw_new(
                "java/lang/IllegalArgumentException",
                "message",
            )
            .expect("Failed to throw exception");
```
 everytime in order to be sure java will catch this cleanly imo

matkt: as it's not related to this PR I think we can merge it

matkt (approved):
LGTM

Issue #130, opened 2023-12-01
Use uncompressed serialisation in ipa_multipoint


thomas-quadratic:
Currently, ipa_multipoint uses compressed serialisation for commitments. The problem is that deserialisation is costly. This issue proposes to use uncompressed serialisation instead as in geth. Although it doubles the size of the serialised commitment, it makes deserialisation much faster.

### Rationale
Commitment is a point on an elliptic curve. In compressed serialisation, we use a byte representation of the field element  x*sign(y) in affine coordinates. This is 32-bytes. However, to deserialise, we need to reconstruct the y-coordinate, and this requires many field operations including a division.

Uncompressed serialisation on the other hand concatenates both x and y. This is 64-bytes now, but deserialisation is trivial.

The geth team has made benchmarks and have chosen uncompressed serialisation. Indeed, Verkle is already much more memory efficient than Merkle, so it was deemed acceptable to use more memory for serialisation given the computation savings.

We propose to follow suit.

### Note
In our case, we use rust implementation that ultimately uses arkworks implementation of elliptic curves. If geth works in projective coordinates, it seems that arkworks uses extended twisted Edwards coordinates. It might or might not have an impact on serialisation.

Issue #128, opened 2023-11-24
Commit should serialise the commitment


thomas-quadratic:
### Description
Currently, commit exposes to the Java-side the computation of a commitment. However, there are several ways to convert a commitment (elliptic curve point in extended twisted-Edwards coordinates) to bytes:

Via map_to_scalars : scalar field element used as values in the parent commitment and the rootHash. It is a relatively expensive operation (division), and not necessarily invertible.
Via to_bytes: base field element used to uniquely represent the commitment. It is cheap and invertible: an equivalent element in banderwagon can be recovered from the base field element.
Via uncompressed serialization: in geth, uncompressed serialization is used to save commitment to the database. This should strike a balance computation and memory. It might use 2 coordinates instead of a single one.

For our application, we actually need to be able to retrieve both the commitment as an elliptic point for proof generation and update commitments, and the commitment_hash to commit a commitment as a value to its parent node.
Although hash can be recovered from the point, we probably prefer to store both to avoid further divisions.

PR #146, opened 2024-02-01
add groupToField and return uncompressed serialized commitments

dragan2234:
Switching to uncompressed commitments and adding a groupToField method.

Closes https://github.com/hyperledger/besu-native/issues/130 
Closes https://github.com/hyperledger/besu-native/issues/128

This is paired with https://github.com/hyperledger/besu-verkle-trie/pull/37


matkt (approved):
LGTM

PR #150, opened 2024-02-12
fix native-build.sh update rust version

dragan2234:
quick fix @matkt


matkt (approved):
LGTM, thanks for the fix

PR #151, opened 2024-02-18
chore!: Modify functionality of `commitRoot` to use `groupToField`

kevaundray:
## Rationale

For the Java library, this will eventually reduce the API surface needed. `commitRoot` will essentially become a call to `commitToScalar` + `groupToField`.

This issue expands on the rationale: https://github.com/crate-crypto/rust-verkle/issues/86

---

There are some geth and go-verkle PRs which would be good to merge before merging this (Leaving this as a Draft while those are open)

PR #152, opened 2024-02-26
Add prover and verifier calls

dragan2234:
Will wait for https://github.com/hyperledger/besu-native/pull/151 and/or changes to pedersenHash before asking for review on this


matkt:
will close it for the moment .  we will see if we want to open it again when we will implement the proof creation @thomas-quadratic

PR #156, opened 2024-03-07
chore!: Updates rust-verkle dependency

kevaundray:
updates the rust verkle dependency to the latest version and modifies the code to match.

- PedersenHash has changed internally to use MapToScalarField
- Methods now take a context and not a committer to make calling them easier.


kevaundray:
supercedes https://github.com/hyperledger/besu-native/pull/151

PR #157, opened 2024-03-07
chore!: Removes pedersenHash method

kevaundray:
Once https://github.com/hyperledger/besu-verkle-trie/pull/43 is merged, the `pedersenHash` method will no longer be needed.

This PR removes it from the API.


matkt (approved):
LGTM

PR #158, opened 2024-03-08
Error handling and init no-copy in JNI

thomas-quadratic:
This PR has two purposes:

1. Error handling: as much as possible, pub extern fn should not panic. We catch errors and throw exceptions in the JVM.
2. No-copy: as much as possible, we should not have to copy data before passing to the FFI.

This is still a draft PR for the moment, as I am making sure tests are in sync.


ipa-multipoint/ipa_multipoint_jni/Cargo.toml
```diff
@@ -22,7 +22,7 @@ repository = "https://github.com/hyperledger/besu-native"
 edition = "2018"
 
 [dependencies]
-ffi_interface = { git = "https://github.com/crate-crypto/rust-verkle", rev = "7688f0aedfb147d3d391abfe8495e46c46d72ce0" }
+ffi_interface = { git = "https://github.com/crate-crypto/rust-verkle", branch = "master" }
```
matkt, line 25: I prefer to have a specific commit in order to build the same thing everytime

matkt (commented):
LGTM, if we can just add a specific commit for rust verkle and fix javadoc issues before merging

It will be nice to add some description for each method in order to simplify the readability

ipa-multipoint/ipa_multipoint_jni/Cargo.toml
thomas-quadratic: No problem, done.

matkt (approved):

PR #182, opened 2024-06-26
0.9.0 errata, move to tmio for tuweni dep

garyschulte:
Remove tuweni as a runtime dependency for gnark-crypto, back to testImplementation dependency.  Also, bump tuweni dep to tmio.

In besu we have to specifically exclude this dependency from the gnark dep to prevent conflicts with besu's version of tuweni.

### DIFF
--- a/gnark/src/main/java/org/hyperledger/besu/nativelib/gnark/LibGnarkEIP196.java
+++ b/gnark/src/main/java/org/hyperledger/besu/nativelib/gnark/LibGnarkEIP196.java
@@ -68,16 +68,6 @@ public class LibGnarkEIP196 {
     return ret;
   }
 
-  public int findFirstTrailingZeroIndex(byte[] array) {
-    for (int i = array.length - 1; i >= 0; i--) {
-      if (array[i] != 0) {
-        return i + 1; // The first trailing zero is after this non-zero byte
-      }
-    }
-    // If all bytes are zero, return 0
-    return 0;
-  }
-
   public static native int eip196altbn128G1Add(
       byte[] input,
       byte[] output,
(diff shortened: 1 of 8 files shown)


gnark/src/main/java/org/hyperledger/besu/nativelib/gnark/LibGnarkEIP196.java
```diff
@@ -61,15 +60,23 @@ public static int eip196_perform_operation(
...
-      err_len.setValue(errBytes.size() - errBytes.numberOfTrailingZeroBytes());
+      err_len.setValue(LibGnarkUtils.findFirstTrailingZeroIndex(err));
       o_len.setValue(0);
     } else {
       err_len.setValue(0);
     }
     return ret;
   }
 
+  public int findFirstTrailingZeroIndex(byte[] array) {
```
pinges, line 71: this is implemented in the Utils class. I think this is not needed ...
You are even using the Utils implementation further up :-)

pinges (changes requested):
see comment

gnark/src/main/java/org/hyperledger/besu/nativelib/gnark/LibGnarkEIP196.java
```diff
@@ -61,15 +60,23 @@ public static int eip196_perform_operation(
...
+    for (int i = array.length - 1; i >= 0; i--) {
+      if (array[i] != 0) {
+        return i + 1; // The first trailing zero is after this non-zero byte
+      }
+    }
+    // If all bytes are zero, return 0
+    return 0;
+  }
```
pinges, line 79:
```suggestion
```

gnark/src/main/java/org/hyperledger/besu/nativelib/gnark/LibGnarkEIP196.java
garyschulte: where?  would love to DRY this out, but I don't see the usage you are talking about

garyschulte: found and removed ;). thx

jflo: This implementation has no usages, and can now be removed.

jflo (approved):
suggest removing removed method via removal

PR #184, opened 2024-06-27
Constantine bindings for EIP196

NickSneo:
Fixes
https://github.com/hyperledger/besu/issues/7242
https://github.com/hyperledger/besu-native/issues/195
https://github.com/hyperledger/besu/issues/7243


garyschulte:
👀 

just a tracking comment

mratsim (commented):
There shouldn't be a need for indirection and allocation, the function can be called and write directly in the relevant buffer, see C, Rust and Go code as referemce.

Also is optimizing for ARM desirable?

On non-x86 Constantine compiles through uint128 as it's the most portable and best perf across all compilers. Both GCC and Clang support an `add with overflow` called `__builtin__addcll` but the perf is abysmal on GCC: https://gcc.godbolt.org/z/jdecvffaP see overall issue https://github.com/mratsim/constantine/issues/357

I do have a plan for optimizing for ARM targets (see https://github.com/mratsim/constantine/issues/200 ) but so far no time to tackle it.

cc @Vindaar as well as he is currently maintaining the external languages dependencies.

constantine/constantine_eip196_bindings.nim
```diff
@@ -0,0 +1,34 @@
...
+    cttEVM_PointNotOnCurve
+    cttEVM_PointNotInSubgroup
+
+
+proc eth_evm_bn254_g1add*(r: ptr uint8, rLen: cint, inputs: ptr uint8, inputsLen: cint): CttEVMStatus {.exportc.} =
+  var result: seq[byte] = newSeq[byte](rLen)
+  var inputSeq: seq[byte] = cast[seq[byte]](inputs[0 ..< inputsLen])
+  let status = ethereum_evm_precompiles.eth_evm_bn254_g1add(result, inputSeq)
+  copyMem(r, addr result[0], rLen)
+  return status
```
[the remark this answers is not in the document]
NickSneo: Hey @mratsim Thanks for the inputs, I have been working on a new approach based on your inputs, taking inspiration from example c code.
ethereum_evm_precompiles.nim -> compile into shared library -> Create C Bindings: ethereum_evm_precompiles.c  -> Define Java Native Methods -> Compile and Run

NickSneo:
@garyschulte @mratsim Pushed some new commits, I shall be removing the files inside constantine/constantine-jni/lib/ folder
and planning to compile constantine lib using gradle task and use those files directly.
The PR is still in draft but can you review to check my approach is I am going in correct direction.

Thanks

constantine/build.gradle
```diff
@@ -0,0 +1,166 @@
...
+task linuxLibCopy(type: Copy) {
+    from 'constantine-jni/lib/libconstantine.so'
+    into 'build/resources/main/linux-x86-64'
+}
+
+task linuxArm64LibCopy(type: Copy) {
+    from 'constantine-jni/lib/libconstantine.so'
+    into 'build/resources/main/linux-aarch64'
+}
+
```
mratsim, line 77: And you probably need a .dll for Windows, if it's one of the platforms you support.

constantine/constantine-jni/lib/libconstantine.a
mratsim, line 1: I don't think those are expected to be checked out do they? There probably should be a .gitignore here

constantine/constantine-jni/src/main/c/ethereum_evm_precompiles.c
```diff
@@ -0,0 +1,69 @@
...
+JNIEXPORT jint JNICALL Java_Constantine_ctt_1eth_1evm_1bn254_1g1add(JNIEnv *env, jobject obj, jbyteArray jr, jint r_len, jbyteArray jinputs, jint inputs_len) {
+    jbyte *r = (*env)->GetByteArrayElements(env, jr, NULL);
+    jbyte *inputs = (*env)->GetByteArrayElements(env, jinputs, NULL);
+
+    ctt_evm_status status = ctt_eth_evm_bn254_g1add((byte *)r, (ptrdiff_t)r_len, (const byte *)inputs, (ptrdiff_t)inputs_len);
+
+    if (status != cttEVM_Success) {
+        printf("ctt_eth_evm_bn254_g1add failed with status: %d\n", status);
+    } else {
+        printByteArray("Result", (const byte *)r, r_len);
```
mratsim, line 26: I assume this is WIP for debugging?

constantine/constantine-jni/src/main/c/ethereum_evm_precompiles.c
```diff
@@ -0,0 +1,69 @@
...
+        if (i < len - 1) {
+            printf(", ");
+        }
+    }
+    printf("]\n");
+}
+
+JNIEXPORT jint JNICALL Java_Constantine_ctt_1eth_1evm_1bn254_1g1add(JNIEnv *env, jobject obj, jbyteArray jr, jint r_len, jbyteArray jinputs, jint inputs_len) {
+    jbyte *r = (*env)->GetByteArrayElements(env, jr, NULL);
+    jbyte *inputs = (*env)->GetByteArrayElements(env, jinputs, NULL);
```
mratsim, line 19: Unfortunately I'm not familiar with Java at all but it seems like what is also being done in c-kzg-4844 so it's likely the expected way to wrap:

https://github.com/ethereum/c-kzg-4844/blob/1bccee0878ffc80efe8741afdb5793ef9105aa35/bindings/java/c_kzg_4844_jni.c#L148-L179

constantine/constantine-jni/lib/libconstantine.a
NickSneo: sure, will add them to gitignore. 
Question - for generating shared lib should we include `nimble make_lib` command in gradle.plugin ? and then copy those to our constantine-jni directory?

constantine/constantine-jni/src/main/c/ethereum_evm_precompiles.c
NickSneo: Yes will remove these print statements

constantine/build.gradle
NickSneo: @garyschulte We need for windows support too?

constantine/constantine-jni/lib/libconstantine.a
mratsim: I think so, this way it stays up-to-date with the compile flags I use.

constantine/build.gradle
garyschulte: windows builds of besu-native is on our radar,  but we do not currently build other besu-native libs for windows.  Having Constantine build for windows would be a nice-to-have for sure.

constantine/constantine-jni/src/main/c/ethereum_evm_precompiles.c
```diff
@@ -0,0 +1,69 @@
...
+    (*env)->ReleaseByteArrayElements(env, jinputs, inputs, 0);
+
+    return (jint)status;
+}
+
+JNIEXPORT jint JNICALL Java_Constantine_ctt_1eth_1evm_1bn254_1g1mul(JNIEnv *env, jobject obj, jbyteArray jr, jint r_len, jbyteArray jinputs, jint inputs_len) {
+    jbyte *r = (*env)->GetByteArrayElements(env, jr, NULL);
+    jbyte *inputs = (*env)->GetByteArrayElements(env, jinputs, NULL);
+
+    ctt_evm_status status = ctt_eth_evm_bn254_g1mul((byte *)r, (ptrdiff_t)r_len, (const byte *)inputs, (ptrdiff_t)inputs_len);
```
garyschulte, line 39: very tidy that constantine already has precompile friendly endpoints 👍

garyschulte (commented):
This PR is looking good.  A good next step would be to implement the EIP196 test cases that exist for the matter-labs and gnark-crypto implementations.  

Much of the contract 'shim' [implementation](https://github.com/hyperledger/besu-native/tree/main/gnark/src/test/java/org/hyperledger/besu/nativelib/gnark) for the besu evm can be copied wholesale, including the [test cases](https://github.com/hyperledger/besu-native/tree/main/gnark/src/test/resources/org/hyperledger/besu/nativelib/gnark).  


Edit: looking at the CI output, it looks like there is a problem loading the generated library in java.

constantine/build.gradle
```diff
@@ -0,0 +1,166 @@
...
+task generateJNIHeader(type: Exec) {
+    description = 'Generates the JNI header file'
+    commandLine 'javac', '-h', 'constantine/constantine-jni/include', 'src/main/java/org/hyperledger/besu/nativelib/constantine/LibConstantineEIP196.java'
+}
+
+task compileNativeLibrary(type: Exec) {
+    description = 'Compiles the native library'
+    commandLine 'clang', '-I', "${System.env.JAVA_HOME}/include", '-I', "${System.env.JAVA_HOME}/include/darwin", '-shared', '-o', 'constantine-jni/lib/libconstantine.jnilib', 'constantine-jni/src/main/c/ethereum_evm_precompiles.c', '-I', 'constantine/include', '-I', 'src/main/java/org/hyperledger/besu/nativelib/constantine', '-L', 'constantine-jni/lib', '-lconstantine'
+}
```
garyschulte, line 42: Idiomatic besu-native builds incorporate the build of the library into the root `build.sh` script and use automake or a similar tool to ensure that the appropriate toolchain is configured (cc, gcc, clang, etc).  

I think it is reasonable to use gradle for these tasks, but we should expect to be able to build on a variety of platforms.

garyschulte: also, adding a `dependsOn` chain between these dependent tasks will make for an easier build ux.

constantine/constantine-jni/src/main/c/ethereum_evm_precompiles.c
```diff
@@ -0,0 +1,69 @@
+#include <jni.h>
+#include "org_hyperledger_besu_nativelib_constantine_LibConstantineEIP196.h"
```
garyschulte, line 2: where is this generated?

constantine/build.gradle
garyschulte, line 42:
```suggestion
task makeConstantine(type: Exec) {
    description = "build constantine source"
    workingDir = file('./constantine')
    commandLine 'nimble', 'make_lib'
}

task compileNativeLibrary(type: Exec) {
    description = 'Compiles the native library'
    dependsOn generateJNIHeader, makeConstantine
    commandLine 'clang', '-I', "${System.env.JAVA_HOME}/include", '-I', "${System.env.JAVA_HOME}/include/darwin", '-shared', '-o', 'constantine-jni/lib/libconstantine.jnilib', 'constantine-jni/src/main/c/ethereum_evm_precompiles.c', '-I', 'constantine/include', '-I', 'src/main/java/org/hyperledger/besu/nativelib/constantine', '-I', 'constantine/constantine-jni/include', '-L', 'constantine-jni/lib', '-lconstantine'
}
```

garyschulte (commented):
generally LGTM.  some suggestions in comments.  It would be helpful to add a README.md that describes the library, purpose, and build instructions (like installing nim, clang, etc as build dependencies).

The changes I suggested to build.gradle plus adding `nim` to the package installs [here](https://github.com/hyperledger/besu-native/blob/main/.github/workflows/build.yml#L20), [here](https://github.com/hyperledger/besu-native/blob/main/.github/workflows/build.yml#L121), and [here](https://github.com/hyperledger/besu-native/blob/main/.github/workflows/build.yml#L187) portion of .github/workflows/build.yml should get things building properly in CI

constantine/constantine-jni/src/main/c/ethereum_evm_precompiles.c
mratsim: Yes, and SHA256 and MODEXP. The plan is to implement all cryptographic precompiles.

For the record, Constantine was actually started with the idea of having a clean from scratch BN254 backend for Nim but it was too hard/time-consuming at the time so a 1-1 port of https://github.com/zcash-hackworks/bn was made instead (https://github.com/status-im/nim-bncurve).
However performance is bad as even a simple add with carry instruction is transformed into 7 instructions: https://github.com/zcash-hackworks/bn/blob/master/src/arith.rs#L396-L405

The issue of zcash's `bn` were already mentioned in 2008 in EIP-1108 https://eips.ethereum.org/EIPS/eip-1108 that reduced gas cost price. And moving to Cloudflare bn256 improved speed by 10x:

> and corroborated in [a later note](https://github.com/ethereum/go-ethereum/pull/16301#issuecomment-372687543), the computational cost of ECADD, ECMUL, and pairing checks (excepting the constant) has dropped roughly an order of magnitude across the board.

However that Cloudflare library was about 2x slower than state-of-the-art in 2021 https://hackmd.io/@gnark/eccbench
https://user-images.githubusercontent.com/22738317/162632923-b057c0a2-6929-419e-b0de-4fd187a5e508.png
and Constantine there had a parameter passing bug which made it 2x slower than necessary on pairings (https://github.com/nim-lang/Nim/issues/16897)

See full context in https://ethresear.ch/t/releasing-constantine-v0-1-0-a-modular-cryptography-stack-for-ethereum/19990

constantine/build.gradle
NickSneo: Okay, I will skip it for now for windows, until we do the performance comparison tests

NickSneo:
> This PR is looking good. A good next step would be to implement the EIP196 test cases that exist for the matter-labs and gnark-crypto implementations.
> 
> Much of the contract 'shim' [implementation](https://github.com/hyperledger/besu-native/tree/main/gnark/src/test/java/org/hyperledger/besu/nativelib/gnark) for the besu evm can be copied wholesale, including the [test cases](https://github.com/hyperledger/besu-native/tree/main/gnark/src/test/resources/org/hyperledger/besu/nativelib/gnark).
> 
> Edit: looking at the CI output, it looks like there is a problem loading the generated library in java.

Fixing it right now

constantine/build.gradle
NickSneo: Added this in my new commit

NickSneo:
> generally LGTM. some suggestions in comments. It would be helpful to add a README.md that describes the library, purpose, and build instructions (like installing nim, clang, etc as build dependencies).
> 
> The changes I suggested to build.gradle plus adding `nim` to the package installs [here](https://github.com/hyperledger/besu-native/blob/main/.github/workflows/build.yml#L20), [here](https://github.com/hyperledger/besu-native/blob/main/.github/workflows/build.yml#L121), and [here](https://github.com/hyperledger/besu-native/blob/main/.github/workflows/build.yml#L187) portions of .github/workflows/build.yml should get things building properly in CI

This and adding tests from Gnark and matter-labs is remaining, working on them.

NickSneo:
@garyschulte Hey, seems the CI pipeline is failing, I tried to fix it but seems nim and nimble installed during the jobs are not available during final-assembly. Can you help me in finding out the issue? 
Is it due to I am doing make and compile using build.gradle and I should switch to build.sh for this compiling and make lib thing?

.github/workflows/build.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           # rust dependencies
           export CARGO_HOME="$HOME/.cargo"
           curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.75.0
+          # nim dependencies
+          export CHOOSENIM_CHOOSE_VERSION=2.0.4
+          curl https://nim-lang.org/choosenim/init.sh -sSf | sh -s -- -y
+          export PATH=$HOME/.nimble/bin:$PATH
```
garyschulte, line 34: if we change the image to ubuntu-23.10 , we can install nim from apt.  Looks like it is version 1.16.14.  So maybe downloading from nim-lang is a good idea.  the curl-scripted kind of installation is a supply chain attack vector though.  Same could be said for the installation of rust right above it 🤔

constantine/build.gradle
```diff
@@ -0,0 +1,171 @@
...
+        languageVersion = JavaLanguageVersion.of(21)
+    }
+}
+
+task makeConstantineLib(type: Exec) {
+    description = 'Compiles the Constantine library'
+    workingDir = file('./constantine')
+    environment 'PATH', "${System.env.PATH}:${System.getenv('HOME')}/.nimble/bin"
+    environment 'CTT_LTO', 'false'
+    commandLine "${System.getenv('HOME')}/.nimble/bin/nimble", 'make_lib'
```
garyschulte, line 29:
```suggestion
    commandLine "nimble", 'make_lib'
    inputs.dir file('./constantine/constantine')
    outputs.dir file('./constantine/lib')```
    
    gradle needs to know about inputs and outputs to reason about if/what to cache.  What is odd is I didn't have the same problem locally.  Removing the explicit path to nim (since you already added it to PATH) makes this build cleanly for me on my local

File: constantine/build.gradle [hunk 10]
```diff
@@ -0,0 +1,171 @@
...
+    jvmArgs = ['-Djava.library.path=out/org/hyperledger/besu/nativelib/constantine']
+    dependsOn compileNativeLibrary
+}
+
+tasks.withType(JavaCompile) {
+    options.encoding = 'UTF-8'
+}
+
+task macArmLibCopy(type: Copy) {
+    from 'constantine/lib/libconstantine.dylib'
```
garyschulte, line 70:
```suggestion
    from 'constantine/lib/libconstantine.dylib'
        dependsOn compileNativeLibrary
```

constantine/build.gradle
```diff
@@ -0,0 +1,171 @@
...
+    into 'build/resources/main/darwin-aarch64'
+}
+
+task macLibCopy(type: Copy) {
+    from 'constantine/lib/libconstantine.dylib'
```
garyschulte, line 75:
```suggestion
    dependsOn compileNativeLibrary
    from 'constantine/lib/libconstantine.dylib'
```

constantine/build.gradle
```diff
@@ -0,0 +1,171 @@
...
+    into 'build/resources/main/darwin-x86-64'
+}
+
+task linuxLibCopy(type: Copy) {
+    from 'constantine/lib/libconstantine.so'
```
garyschulte, line 80:
```suggestion
    dependsOn compileNativeLibrary
    from 'constantine/lib/libconstantine.so'
```

constantine/build.gradle
```diff
@@ -0,0 +1,171 @@
...
+    into 'build/resources/main/linux-x86-64'
+}
+
+task linuxArm64LibCopy(type: Copy) {
+    from 'constantine/lib/libconstantine.so'
```
garyschulte, line 85:
```suggestion
    dependsOn compileNativeLibrary
    from 'constantine/lib/libconstantine.so'
```

.github/workflows/build.yml
NickSneo: The problem is in the final-assembly step, because I checked this step is working fine and installing nim correctly.
I see steps like these in final-assembly 
```
      - name: Download bls12-381
        uses: actions/download-artifact@v3
        with:
          name: bls12-381 native build artifacts
          path: bls12-381/build/
```

I checked building with nim 2.0.4 which worked fine for me.
In constantine lib - `Constantine can be compiled by Nim v1.6.x, v2.0.2 and v2.0.4 but not Nim v2.0.0 (due to a compile-time evaluation crash)` 
But installing with apt won't change much as even choose-nim is working fine

garyschulte: Yup, I would expect it will work fine, but giving the contents of a URL the privilege to run anything in the shell is a really sketchy way to install dependencies.  We are doing the same thing for rust, and we really shouldn't.

garyschulte:
I repeated naive testing I did when comparing matter-labs, blst, and gnark-crypto.  This PR compares favorably at least on this test machine:

Screenshot 2024-07-19 at 12 52 08 PM
Screenshot 2024-07-19 at 12 52 22 PM
Screenshot 2024-07-19 at 12 52 15 PM

NickSneo:
This looks good, performance seems similar to Gnark but in G1Mul Gnark seems better. Should we go ahead with EIP2537?

garyschulte:
> This looks good, performance seems similar to Gnark but in G1Mul Gnark seems better. Should we go ahead with EIP2537?

👍 I think yes, these are good results that warrant adding bls12-381 also.  Probably in a follow-on PR to keep the review scope down.  I will :eyes: on the CI build issue

NickSneo:
> > This looks good, performance seems similar to Gnark but in G1Mul Gnark seems better. Should we go ahead with EIP2537?
> 
> 👍 I think yes, these are good results that warrant adding bls12-381 also. Probably in a follow-on PR to keep the review scope down. I will 👀 on the CI build issue

Sure, I will also try to fix it, maybe just missing some small part. Thanks!

garyschulte:
> Sure, I will also try to fix it, maybe just missing some small part. Thanks!

I will defer to you, happy to help if you need unblocking.

.github/workflows/build.yml
mratsim: 1.6.x (latest is 1.6.20) works fine for Constantine and it's tested in CI. No benefits performance-wise for latest Nim versions. Constantine uses low-level Nim and has no library dependencies and even no std-lib dependencies or allocator dependencies to be extremely hard to supply-chain attack (i.e. introducing a compiler bug).
Hence the features that might justify upgrading a compiler are more related to its compile-time capabilities and type system which Constantine uses extensively.

mratsim:
> This looks good, performance seems similar to Gnark but in G1Mul Gnark seems better. Should we go ahead with EIP2537?

I'm curious how I can reproduce this bench, on my machine raw perf (i.e. not the EVM wrapper), constantine is faster than Gnark by 15%

image
image

garyschulte:
> I'm curious how I can reproduce this bench, on my machine raw perf (i.e. not the EVM wrapper), constantine is faster than Gnark by 15%

it is a naive performance comparison, just running the unit test cases repeatedly 1000x for each input in a tight loop (only 100x for pairing test).

garyschulte:
@NickSneo do you want or need help with the CI troubles?  I think this PR is very promising, especially in light of the fact that we would like to have Constantine verkle bindings in besu-native.  

IMO besu-native would benefit from having generalized bindings for both gnark-crypto and Constantine, rather than a collection of bespoke libraries.  In that regard, I am keenly interested in getting this PR merged to get Constantine into the repo.

garyschulte:
> > I'm curious how I can reproduce this bench, on my machine raw perf (i.e. not the EVM wrapper), constantine is faster than Gnark by 15%
> 
> it is also worth noting that we are not using a release version of gnark-crypto, but rather built from [this sha](https://github.com/Consensys/gnark-crypto/commit/0823e43febdc)

Sorry - somehow I edited your comment rather than replying.

NickSneo:
@garyschulte Hey, with latest commit, everything is working fine with correct build.sh compiling nim shared lib and then in gradle.build copy it to correct place and tests working fine. But Don't why it is failing in pipeline

garyschulte:
The problem is the way CI expects to build these libraries.  It expects to run `./build.sh` on four different architectures to build the corresponding shared library - once each for x86 and arm64 versions of both linux and mac.  The result of each build step should be the upload of the shared library(or libraries) as an artifact with the operating system and arch as part of the path inside the artifact.  Then on the final assembly step it expects each subproject to take each of the native artifacts and build a jar with the java code and the native artifacts embedded in it.

In the final product, the process expects to be able to find the four native artifacts from each of the four build steps so it can embed them into the final multi-arch jar:

darwin-aarch64/libconstantine.dylib
darwin-x86-64/libconstantine.dylib
linux-x86-64/libconstantine.so
linux-aarch64/lib/libconstantine.so

Since CI is expecting to build across architectures, we should probably put a constantine build step into build.sh like:
```
build_constantine() {
  cat <<EOF
  ############################
  ####### build constantine #######
  ############################
EOF

  cd "$SCRIPTDIR"

  # delete old build dir, if exists
  rm -rf "$SCRIPTDIR/constantine/build" || true
  mkdir -p "$SCRIPTDIR/constantine/build/lib"

  if [[ "$OSTYPE" == "msys" ]]; then
    	LIBRARY_EXTENSION=dll
  elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
    LIBRARY_EXTENSION=so
  elif [[ "$OSTYPE" == "darwin"* ]]; then
    LIBRARY_EXTENSION=dylib
  fi

  ./gradlew :constantine:clean :constantine:compileNativeLibrary

  mkdir -p "$SCRIPTDIR/constantine/build/${OSARCH}/lib"
  cp constantine/constantine/lib/libconstantine.* "$SCRIPTDIR/constantine/build/${OSARCH}/lib"
}

...

build_constantine
```

and make the requisite changes to the jar building process for the constantine subproject so that it pulls in the jars from the os-arch friendly artifact directories.

Lastly, we need to add upload and download steps for constantine artifacts in build.yml native build tasks.  In each native build task, add an additional step like:
```
      - uses: actions/upload-artifact@v3.1.0
        with:
          name: constantine native build artifacts
          path: constantine/constantine/lib
```

and in the final-assembly task, pull all of the artifacts down by adding a step like:
```
      - name: Download constantine
        uses: actions/download-artifact@v3
        with:
          name: constantine native build artifacts
          path: constantine/constantine/lib
```

NickSneo:
Hey @garyschulte , I was able to fix the pipeline issue.
I was using my work laptop which is Mac (m2) for testing and everything was working fine, but since the pipeline was failing in ubuntu x86_64, so I dual booted my PC with ubuntu and tested it in there. Seems there was an issue with `LD_LIBRARY_PATH` for which I added the required fix.

Now pipeline is passing. Please review and let me know if anything else is required from my side in this PR.

build.sh
```diff
@@ -319,13 +319,53 @@ EOF
...
+build_constantine() {
+  echo "#############################"
+  echo "####### build constantine ####"
+  echo "#############################"
+
+  # Skip if OSARCH is linux-gnu-aarch64
+  if [[ "$OSARCH" == "linux-gnu-aarch64" ]]; then
+    echo "Skipping build for OSARCH: ${OSARCH}"
+    return
+  fi
```
garyschulte, line 331: is this a CI-specific workaround?  it looks like an arm64 linux host env would not be able to build constantine?

garyschulte (approved):
This LGTM 👍 

One non-blocking question about linux arm64 builds though.

build.sh
NickSneo: I skipped Linux arm64 build since nim installer was not compatible with it, and I didn't have an Linux arm environment to test it.
Will test and add in the follow on PR

garyschulte:
Merging as-is, with an expectation that linux-arm64 will be added in a follow-on PR, as the CHOOSENIM install method does not work on linux arm64:
```
choosenim-init: Error: Sorry, your platform (linux_arm64) is not supported by choosenim.
choosenim-init: Error: You will need to install Nim using an alternative method.
choosenim-init: Error: See the following link for more info: https://nim-lang.org/install.html
```

PR #185, opened 2024-06-27
Bugfix/eip 196 edge case

garyschulte:
fix for a weird (imo) EIP-196 expectation that a malformed/incomplete g1 point should return an error if the X value is readable, but not in the field; whereas a malformed/incomplete g1 point will otherwise return 0


gnark/gnark-jni/gnark-eip-196.go
```diff
@@ -100,7 +121,17 @@ func eip196altbn128G1Mul(javaInputBuf, javaOutputBuf, javaErrorBuf *C.char, cInp
     errorBuf := castErrorBuffer(javaErrorBuf, errorLen)
 
     if inputLen < EIP196PreallocateForG1 {
-        // if we do not have complete input, return 0
+
+        // check the X point and return an error if it is in the field, edge case
```
pinges, line 125:
```suggestion
        // check the X point and return an error if it is not in the field, edge case
```

gnark/gnark-jni/gnark-eip-196.go
```diff
@@ -38,29 +47,37 @@ func eip196altbn128G1Add(javaInputBuf, javaOutputBuf, javaErrorBuf *C.char, cInp
...
-        return 0
-    }
-
     // Convert error C pointers to Go slices
     errorBuf := castErrorBuffer(javaErrorBuf, errorLen)
 
+    // check we have input size sufficient for a G1Affine
+    if inputLen < EIP196PreallocateForG1 {
+
+        // if not, check the X point and return an error if it is in the field, edge case
```
pinges, line 61:
```suggestion
        // if not, check the X point and return an error if it is not in the field, edge case
```

pinges (commented):
I'm not sure about the edge case. For me it does not make sense and we should try to find out whether that is the right behaviour.

garyschulte:
> I'm not sure about the edge case. For me it does not make sense and we should try to find out whether that is the right behaviour.

This PR to resolve a reference test failure.  I agree it is weird behavior.  IMO all malformed inputs should return an error.

pinges (approved):
LGTM

jflo (approved):
I don't understand the math, but the change in error case handling is sound.

PR #188, opened 2024-07-10
gnark-crypto, eip-196: malformed input handling

garyschulte:
ECADD and ECMUL:
* Treat short input as an error conditions on a per-parameter basis
* Add test cases for point not on curve and point not in field for ECADD
* pass output and error lengths for eip196 rather than static lengths

This change is consistent with reference tests and parameter parsing in other crypto implementations, and clarifies the edge cases (and language) described in the original eip:

https://eips.ethereum.org/EIPS/eip-196#test-cases

using this pr, clean run of [besu#7626](https://github.com/hyperledger/besu/pull/7262) :
* referenceTest task
* evmtool based exeution-spec-test main


pinges (approved):
LGTM. I like that we are getting rid of that edge case :-)

PR #196, opened 2024-07-29
Check released version and configured version

cdivitotawela:
Recently same version published and caused issues for the downstream applications with checksum changes. Added GitHub step check whether the version configured has been released previously. This will prevent overriding existing releases

fixes consensys/protocol-misc#937

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
             echo "Query for version [$VERSION]"
             curl --head -L --silent --output /dev/null --fail "$ART_URL/$VERSION" &&  {
               echo "ERROR: Binary arithmetic already exists for version [$VERSION]"
-              echo "ERROR: Blocking publishing same version to the Besu Artiactory"
+              echo "ERROR: Blocking republishing same version to the Besu Artifactory"
               echo "url $ART_URL/$VERSION"
               exit 1
             }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
             VERSION=$(grep "^version" gradle.properties | sed 's|^version=\(.*\)$|\1|g')
             echo "Query for version [$VERSION]"
             curl --head -L --silent --output /dev/null --fail "$ART_URL/$VERSION" &&  {
-              echo "ERROR: Binary arithmetic already exit for version [$VERSION]"
+              echo "ERROR: Binary arithmetic already exists for version [$VERSION]"
               echo "ERROR: Blocking publishing same version to the Besu Artiactory"
               echo "url $ART_URL/$VERSION"
               exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -32,6 +32,25 @@ jobs:
...
+      - name: Check existing release
+        run: |
+          # Check whether version contains SNAPSHOT
+          grep "^version" gradle.properties | grep -q SNAPSHOT || {
+            # Extract version from gradle.properties
+            VERSION=$(grep "^version" gradle.properties | sed 's|^version=\(.*\)$|\1|g')
+            echo "Query for version [$VERSION]"
+            curl --head -L --silent --output /dev/null --fail "$ART_URL/$VERSION" &&  {
+              echo "ERROR: Binary arithmetic already exit for version [$VERSION]"
+              echo "ERROR: Blocking publishing same version to the Besu Artiactory"
```
macfarla, line 47:
```suggestion
              echo "ERROR: Blocking republishing same version to the Besu Artifactory"
```

.github/workflows/build.yml
```diff
@@ -32,6 +32,25 @@ jobs:
...
+      # Only check binary for arithmetic as all published together and arithmetic is the first
+      - name: Check existing release
+        run: |
+          # Check whether version contains SNAPSHOT
+          grep "^version" gradle.properties | grep -q SNAPSHOT || {
+            # Extract version from gradle.properties
+            VERSION=$(grep "^version" gradle.properties | sed 's|^version=\(.*\)$|\1|g')
+            echo "Query for version [$VERSION]"
+            curl --head -L --silent --output /dev/null --fail "$ART_URL/$VERSION" &&  {
+              echo "ERROR: Binary arithmetic already exit for version [$VERSION]"
```
macfarla, line 46:
```suggestion
              echo "ERROR: Binary arithmetic already exists for version [$VERSION]"
```

.github/workflows/build.yml
```diff
@@ -32,6 +32,25 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: recursive
+      # Check whether gradle.properties version has been released already
+      # This prevents accidently publish same version
+      # Only check binary for arithmetic as all published together and arithmetic is the first
+      - name: Check existing release
+        run: |
+          # Check whether version contains SNAPSHOT
+          grep "^version" gradle.properties | grep -q SNAPSHOT || {
```
macfarla, line 41: does this mean that republish is allowed for *-SNAPSHOT versions?

macfarla (commented):
@garyschulte can you take a look please

.github/workflows/build.yml
cdivitotawela: Yes. We need to allow publishing *-SNAPSHOT versions.

cdivitotawela: Logic here is that we grep for SNAPSHOT and if not find move into the block for execution.

macfarla (approved):

jframe (approved):

garyschulte:
This seems to have broken CI for releases.  I am going to revert in the 0.9.4 release and we can come up with something else

Issue #198, opened 2024-08-13
Add linux arm64 support in Constantine build


garyschulte:
Initial support for Constantine in besu-native in #184 omitted building for linux-arm64.  Support for linux arm64 is mostly just a matter of getting nim installed in the CI build container for arm64, as the rest of the plumbing already exists.

```
choosenim-init: Error: Sorry, your platform (linux_arm64) is not supported by choosenim.
choosenim-init: Error: You will need to install Nim using an alternative method.
choosenim-init: Error: See the following link for more info: https://nim-lang.org/install.html
```

_Originally posted by @garyschulte in https://github.com/hyperledger/besu-native/issues/184#issuecomment-2287031271_

garyschulte:
It appears that the mac support aarch64 and x86 are not getting included in the final artifact.  The build version for [0.9.4](https://hyperledger.jfrog.io/artifactory/besu-maven/org/hyperledger/besu/constantine/0.9.4/constantine-0.9.4.jar) only includes linux x86 libs:
```
     0 Tue Aug 13 23:09:30 PDT 2024 META-INF/
   246 Tue Aug 13 23:09:30 PDT 2024 META-INF/MANIFEST.MF
     0 Tue Aug 13 23:09:30 PDT 2024 org/
     0 Tue Aug 13 23:09:30 PDT 2024 org/hyperledger/
     0 Tue Aug 13 23:09:30 PDT 2024 org/hyperledger/besu/
     0 Tue Aug 13 23:09:30 PDT 2024 org/hyperledger/besu/nativelib/
     0 Tue Aug 13 23:09:30 PDT 2024 org/hyperledger/besu/nativelib/constantine/
  1939 Tue Aug 13 23:09:30 PDT 2024 org/hyperledger/besu/nativelib/constantine/LibConstantineEIP196.class
     0 Tue Aug 13 23:09:30 PDT 2024 lib/
     0 Tue Aug 13 23:09:30 PDT 2024 lib/linux-gnu-x86_64/
 16608 Tue Aug 13 23:09:30 PDT 2024 lib/linux-gnu-x86_64/libconstantineeip196.so
3116280 Tue Aug 13 23:09:30 PDT 2024 lib/linux-gnu-x86_64/libconstantine.so
```

NickSneo:
[...]

PR #203, opened 2024-08-14
Linux aarch64 support for EIP196 constantine

NickSneo:
Fixes https://github.com/hyperledger/besu-native/issues/198

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -340,6 +340,7 @@ build_constantine() {
     tar -xf nim-2.0.4-linux_arm64.tar.xz
     git config --global --add safe.directory /home/ubuntu/constantine/constantine
     export PATH=$(pwd)/nim-2.0.4/bin:$PATH
+    export NIMFLAGS="-d:release --passC:-fPIC --passL:-fPIC"
     nimble make_lib
   else
     export PATH=$HOME/.nimble/bin:$PATH
--- a/build.sh
+++ b/build.sh
@@ -336,6 +336,7 @@ build_constantine() {
     # Download and extract Nim
     wget https://github.com/nim-lang/nightlies/releases/download/2024-01-04-version-2-0-0acdce502ca39441bb7e4050efdf43bf03db075d/nim-2.0.3-linux_arm64.tar.xz
     tar -xf nim-2.0.3-linux_arm64.tar.xz
+    git config --global --add safe.directory /home/ubuntu/constantine/constantine
     export PATH=$(pwd)/nim-2.0.3/bin:$PATH
     nimble make_lib
   else
--- a/build.sh
+++ b/build.sh
@@ -339,8 +339,6 @@ build_constantine() {
     git config --global --add safe.directory /home/ubuntu/constantine/constantine
     export PATH=$(pwd)/nim-2.0.4/bin:$PATH
     nimble make_lib
-    export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-aarch64
-    export PATH=$JAVA_HOME/bin:$PATH
   else
     export PATH=$HOME/.nimble/bin:$PATH
     nimble make_lib
--- a/build.sh
+++ b/build.sh
@@ -352,7 +352,7 @@ build_constantine() {
    clang -I"${JAVA_HOME}/include" -I"${JAVA_HOME}/include/darwin" -shared -o "$SCRIPTDIR/constantine/build/${OSARCH}/lib/libconstantineeip196.jnilib" ethereum_evm_precompiles.c -Iconstantine/include -I. -Lconstantine/lib -lconstantine
  elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
    cp "$SCRIPTDIR/constantine/constantine/lib/libconstantine.so" "$SCRIPTDIR/constantine/build/${OSARCH}/lib/"
-   clang -I"${JAVA_HOME}/include" -I"${JAVA_HOME}/include/linux" -fPIC -shared -o "$SCRIPTDIR/constantine/build/${OSARCH}/lib/libconstantineeip196.so" ethereum_evm_precompiles.c -Iconstantine/include -I. -Lconstantine/lib -lconstantine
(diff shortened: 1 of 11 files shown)


NickSneo:
@garyschulte Please review.

native-build.sh
```diff
@@ -13,4 +13,8 @@ curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.75.0
 export PATH=$PATH:$CARGO_HOME/bin
 git config --global --add safe.directory /home/ubuntu/secp256r1/besu-native-ec
 git config --global --add safe.directory /home/ubuntu/secp256r1/besu-native-ec/openssl
+wget https://download.java.net/java/GA/jdk21.0.2/f2283984656d49d69e91c558476027ac/13/GPL/openjdk-21.0.2_linux-aarch64_bin.tar.gz
+tar -xzf openjdk-21.0.2_linux-aarch64_bin.tar.gz -C $HOME
+export JAVA_HOME=$HOME/jdk-21.0.2
+export PATH=$JAVA_HOME/bin:$PATH
```
garyschulte, line 19: lets install openjdk-21-jdk from apt instead.  The only place this script is called from is a docker exec command in build.yml.  We should be able to run against 22.04 instead of 20.04 and will have openjdk-21-jdk available without having to do downloads of specific minor revisions

also, we should get rid of openjdk-11 in the apt install

build.sh
```diff
@@ -324,23 +324,25 @@ build_constantine() {
...
   mkdir -p "$SCRIPTDIR/constantine/build/${OSARCH}/lib"
 
-  export PATH=$HOME/.nimble/bin:$PATH
-
   # Build the constantine library
   export CTT_LTO=false
-  nimble make_lib
+  if [[ "$OSARCH" == "linux-gnu-aarch64" ]]; then
+    # Download and extract Nim
+    wget https://github.com/nim-lang/nightlies/releases/download/2024-03-28-version-2-0-b47747d31844c6bd9af4322efe55e24fefea544c/nim-2.0.4-linux_arm64.tar.xz
```
garyschulte, line 337: 👍

garyschulte (approved):
LGTM, minor nit about the the openjdk install source

constantine/build.gradle
```diff
@@ -39,17 +39,36 @@ if (osName.contains('mac') && osArch.contains('aarch64')) {
     libDir = 'linux-gnu-x86_64'
 }
 
-task libCopy(type: Copy) {
-    from "build/${libDir}/lib/"
-    into "build/resources/main/lib/${libDir}"
+task macArmLibCopy(type: Copy) {
+    from "build/darwin-aarch64/lib/libconstantine.dylib"
+    from "build/darwin-aarch64/lib/libconstantineeip196.jnilib"
+    into 'build/resources/main/lib/darwin-aarch64'
```
garyschulte, line 45: FYI, testling locally I am getting
`Exception in thread "main" java.lang.UnsatisfiedLinkError: no constantineeip196 in java.library.path: ...`

NickSneo: Which OS are you testing this in?
I assume first you ran the ./build.sh and then the test files?

I tested in Mac and Linux x86_64, working fine for me

NickSneo: Which OS are you testing this in?
I assume first you ran the ./build.sh and then the test files?

I tested in Mac and Linux x86_64, working fine for me

native-build.sh
NickSneo: Removed openjdk-11 and added openjdk-21

constantine/build.gradle
garyschulte: I am testing by integrating into besu on Darwin aarch64, when I build I am seeing the libs getting embedded in `lib/{arch}` rather than `{arch}`.  The 0.9.4 build shows the same.  

Perhaps there is a classpath difference / expectation in besu-native/constantine that allows it to find the libs, but that same native classpath isn't in besu (since besu is using the jar dep instead of a filesystem build artifact).  Looking into it now

NickSneo:
@garyschulte I haven't changed anything for linux x86_64 but seems the pipeline is failing due to - `/tmp/choosenim-0.8.5_linux_amd64: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/choosenim-0.8.5_linux_amd64)`

Do you have any idea about this?

NickSneo:
@garyschulte Pipeline for Mac Arm is failing due to -

```
 ############################
  ####### build gnark #######
  ############################
./build.sh: line 312: go: command not found
```

garyschulte:
> @garyschulte Pipeline for Mac Arm is failing due to -
> 
> ```
>  ############################
>   ####### build gnark #######
>   ############################
> ./build.sh: line 312: go: command not found
> ```

We can add go, it builds fine locally.  It is just the packaging that is presenting a problem for me at the moment

NickSneo:
@garyschulte Pipelines are passing now.

garyschulte:
Converting to JNA does seem to include a little overhead for the first call to the library, but subsequent performance is essentially identical:

image
image

build.sh
```diff
@@ -332,6 +332,15 @@ build_constantine() {
...
   mkdir -p "$SCRIPTDIR/constantine/build/${OSARCH}/lib"
 
+  # Modify config.nims if on x86_64
+  if [[ "$OSARCH" == "linux-gnu-x86_64" ]]; then
+    {
+      echo 'when defined(linux):'
+      echo '  switch("passC", "-fPIC")'
+      echo '  switch("passL", "-fPIC")'
+    } >> config.nims
+  fi
```
garyschulte, line 342: could we add a check so that we don't append over and over on local builds?
```
# Check and modify config.nims if on x86_64
if [[ "$OSARCH" == "linux-gnu-x86_64" ]]; then
  # Check if the config.nims already contains the necessary flags
  if ! grep -q 'switch("passC", "-fPIC")' config.nims; then
    {
      echo 'when defined(linux):'
      echo '  switch("passC", "-fPIC")'
      echo '  switch("passL", "-fPIC")'
    } >> config.nims
  fi
fi
```

garyschulte:
The weird requirement about darwin vs linux shared library location notwithstanding, lets MERGE this and we can sort out that minor oddity in a subsequent PR 👍

PR #204, opened 2024-08-15
Prevent publishing same version

cdivitotawela:
Workflow checks whether the configured non snapshot version has been published already to prevent re-publish same version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,13 +41,14 @@ jobs:
       # Only check binary for arithmetic as all published together and arithmetic is the first
       - name: Check release published
         run: |
-          # Skip if version contain SNAPSHOT
-          grep "^version" gradle.properties | grep -q SNAPSHOT || {
+          # Only interested in release version
+          if [[ $(grep -E 'version=[0-9]+\.[0-9]+\.[0-9]+$' gradle.properties | wc -l) -ne 0 ]]
+          then
             # Extract version from gradle.properties
             VERSION=$(grep "^version" gradle.properties | sed 's|^version=\(.*\)$|\1|g')
             echo "Query for version [$VERSION]"
             RESPONSE_CODE=$(curl --head -L --silent --output /dev/null  --write-out %{http_code} "$ART_URL/$VERSION")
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -36,6 +36,37 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: recursive
+      # Check whether gradle.properties version has been released already
+      # This prevents accidently publish same version
+      # Only check binary for arithmetic as all published together and arithmetic is the first
+      - name: Check release published
+        run: |
+          # Skip if version contain SNAPSHOT
+          grep "^version" gradle.properties | grep -q SNAPSHOT || {
```
garyschulte, line 45: how about using a negated regex to positively identify a version rather than checking for SNAPSHOT, like 
`grep "^version" gradle.properties | grep -qvE '^version=[0-9]+\.[0-9]+\.[0-9]+$' ||`

I will say that using a conditional if/else statement is more readable than short circuit evaluation though

.github/workflows/build.yml
```diff
@@ -36,6 +36,37 @@ jobs:
...
+            # Extract version from gradle.properties
+            VERSION=$(grep "^version" gradle.properties | sed 's|^version=\(.*\)$|\1|g')
+            echo "Query for version [$VERSION]"
+            RESPONSE_CODE=$(curl --head -L --silent --output /dev/null  --write-out %{http_code} "$ART_URL/$VERSION")
+            echo "DEBUG: Response code from curl command [$RESPONSE_CODE]"
+            if [[ $RESPONSE_CODE == "404" ]]
```
garyschulte, line 51: so this is what we were hung up on previously?  a 404 would mean success, but we were failing because the curl exit code was non zero for a 404?

garyschulte (approved):
LGTM, non-blocking suggestion

.github/workflows/build.yml
cdivitotawela: Yes curl command returned non-zero because of 404 and I had not catch and returned zero return. It was a pure miss

.github/workflows/build.yml
cdivitotawela: Fixes added

PR #205, opened 2024-08-21
debug constantine final assembly

garyschulte:


garyschulte:
closed.  jvm's are just really picky and inconsistent about jna paths

PR #210, opened 2024-09-08
Support for EIP2537 Constantine Lib

NickSneo:
Closes https://github.com/hyperledger/besu/issues/7244


garyschulte (approved):
Looks good, esp performance for g1 and g2 msm precompiles

PR #213, opened 2024-09-24
Update libsecp256k1 to v0.5.1

daniellehrner:


fab-10 (approved):
Looks like there are performance improvements on x86

PR #216, opened 2024-10-01
test data generator / helper for ECC inputs

garyschulte:
helper for test input generation.  Not exhaustive for all ECC precompiles, but we can add as we need.


macfarla (approved):

pinges (approved):
LGTM

PR #217, opened 2024-10-03
add parallel g1/g2 msm gnark-crypto impl

garyschulte:
Implement gnark MultiExp for g1/g2 MSM, using default degree of parallelism.  

image

image

### DIFF
--- a/gnark/gnark-jni/gnark-eip-2537.go
+++ b/gnark/gnark-jni/gnark-eip-2537.go
@@ -170,7 +170,7 @@ func eip2537blsG1MultiExp(javaInputBuf, javaOutputBuf, javaErrorBuf *C.char, cIn
 }
 
 //export eip2537blsG1MultiExpParallel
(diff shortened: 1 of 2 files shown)


siladu (approved):
LGTM.

w.r.t the benchmark results, can you confirm 
1. 24.9.1 is calling `eip2537blsG1MultiExp`
2. besu-gnark-single-parallel is calling `eip2537blsG1MultiExpParallel` but limiting to a single core (by setting nbTasks in the MultExpConfig?) This code isn't in this PR.
3. besu-gnark-parallel is calling `eip2537blsG1MultiExpParallel` using the implementation in this PR.

garyschulte:
> ^ IIUC, then should we make (2) a configurable option, i.e. an option to limit to single core (not sure we need to expose a parameter for arbitrary cores?)

this seems like a reasonable idea.  I think I can add the capacity to set NbTasks in the initialization of the gnark bls lib, and we can plumb an option in besu.  I will 👀 at adding that bit of config and if it is as lightweight as I am thinking, will add it in this PR

PR #235, opened 2025-02-06
revert build to ubuntu 20.04

garyschulte:
20.04 is still supported until May, and there doesn't seem to be any significant benefit to using the more recent glibc in 22.04.

* reverts to 20.04 for CI build


garyschulte:
close, addressing in #236

PR #236, opened 2025-02-06
Fix/migrate to v4 actions

garyschulte:
v3.x actions were [officially unsupported as of Jan 30](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)

This PR:
* Migrates to v4 github actions for upload and download artifacts.  
* reverts to ubuntu 20.04 for linux-x86-64 builds


macfarla (approved):
LGTM

PR #245, opened 2025-03-06
Refactor of library loading to remove platform/arch naming differences

### LINKED ISSUE #242: Use explicit library loading rather than relying on JNA
# Description
JNA's Native.register() function is convenient and tidy, but presents problems on platforms that can have non-standard OS_ARCH names.  For example, some versions of linux might report arch as `linux-x86-64` where others might report that same arch as `linux-gnu-x86_64`.  This causes problems for packaging and auto-loading native libraries across different distributions and operating systems.

besu-native lib loading needs to be more explicit and opinionated in order to ensure native library implementations are available for all supported platforms.  

The c-kzg-4844 java wrapper has [fairly robust](https://github.com/ethereum/c-kzg-4844/blob/252a6b2c2c6fd478bd0ccc0edaf846e5d83f9f73/bindings/java/src/main/java/ethereum/ckzg4844/CKZG4844JNI.java#L16-L21)  native lib loading.  Using a similar strategy would not only be more robust than JNA, but it would also conform the native lib loading behavior across besu's native lib providers.

garyschulte:
Use explict paths based on arch in besu-native during the library builds, and during library loading.  We are still using JNA for jna libraries, and jni for jni libraries.  However we are using conformed arch names to publish and find native libs.  This should resolve the OSTYPE and OSARCH naming problems that seem to plague linux distributions especially.

fixes #242 
is blocking [#8432](https://github.com/hyperledger/besu/issues/8342)

Tested on:
* Mac Os 15.3.1 on M2max
* Ubuntu 24.04 on x86_64
* Pop! OS linux 22.04 x86_64
* Armbian Linux 24.8.4 on riscv64
* Armbian Linux 24.5.5 on arm64

Build fails on:
* Mac Os 10.15.7 on X86 (❌ build fails, OS toolchain is too old to know how to build a lipo lib)

### DIFF
--- a/common/src/main/java/org/hyperledger/besu/nativelib/common/BesuNativeLibraryLoader.java
+++ b/common/src/main/java/org/hyperledger/besu/nativelib/common/BesuNativeLibraryLoader.java
@@ -24,13 +24,10 @@ import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
+import java.util.Arrays;
 import java.util.Optional;
 
 public class BesuNativeLibraryLoader {
-
-  // base path for lib/arch.  replace underscore with dash to conform to jar packaging
-  static final String LIBRARY_RESOURCE_ARCH_PATH =
-      "lib/" + System.getProperty("os.arch").replace("_", "-") + "/";
 
   /**
    * Wraps JNA with a prescriptive path, removing any platform naming inconsistencies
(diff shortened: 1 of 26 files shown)


usmansaleem (approved):
LGTM. My only comment would be to introduce a env file that can be sourced into build.sh and native-build.sh. This can allow single place to maintain resource versions. This can be implement in a future PR though.

common/src/main/java/org/hyperledger/besu/nativelib/common/BesuNativeLibraryLoader.java
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Wraps JNA with a prescriptive path, removing any platform naming inconsistencies
+   *  that might exist.  E.g. linux-gnu-x86-64 vs linux-x86_64.
+   *
+   *  Note that this method expects the directory space to be "partitioned" only by the arch, and
+   *  relies on the different operating systems to have different shared library object
+   *  filenames.  E.g. .dylib vs .so vs .dll.
+   */
+  public static void registerJNA(Class jniClass, String libraryName) {
```
pinges, line 40: should that be Class jnaClass?

common/src/main/java/org/hyperledger/besu/nativelib/common/BesuNativeLibraryLoader.java
```diff
@@ -0,0 +1,126 @@
...
+      String exceptionMessage =
+          String.format(
+              "Couldn't load native library (%s). It wasn't available at %s or the library path.",
+              libraryName, asLibraryResourcePath(libraryName));
+      throw new RuntimeException(exceptionMessage);
+    }
+  }
+
+
+  private static Optional<Path> extract(Class jniClass, String libraryName) {
```
pinges, line 80: This is used for jni and jna, so should we rename jniClass here?

pinges (commented):
LGTM, just maybe change two var names.
Really like that!!!

PR #256, opened 2025-03-17
Update test vectors

daniellehrner:
Use more performant input checks, update test vectors to reflect.

* update to use gofmt


garyschulte (approved):
LGTM

PR #259, opened 2025-03-17
chore: add docs for exposed JNI gnark methods

kevaundray:

PR #260, opened 2025-03-17
chore: split JNI logic from pure golang logic

kevaundray:
This is built ontop of #259 so the diff will include those changes as that PR is still open.

This essentially splits the logic into three parts:

- Validate input and convert to golang data structures (this part is "unsafe" due to interactions with JNI and C)
- Process the input and perform the necessary operation (this will be a golang slice)
- Copy the output into the output buffer (this part is also "unsafe" if the output buffer does not satisfy the JNI requirements)


kevaundray:
> can you squash commit/rebase to fix the DCO requirement?

Should be fixed now :)

PR #261, opened 2025-03-17
chore: replace reflect.SliceHeader

kevaundray:
As of Go 1.17 unsafe.Slice is the recommended way since the go runtime does not make any assumptions about the layout of the slice header


gnark/gnark-jni/gnark-eip-196.go
```diff
@@ -8,8 +8,9 @@ import (
 	"errors"
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 	"math/big"
-	"reflect"
 	"unsafe"
+
+	"github.com/consensys/gnark-crypto/ecc/bn254"
```
garyschulte, line 13: this diff makes the import look like bn254 is imported twice

garyschulte (commented):
LGTM, possibly a duplicate import and will need merge or rebase after #260

PR #265, opened 2025-03-19
chore: Add safety note around zero initializing JavaOutputBuf

kevaundray:


kevaundray:
In `nonMontgomeryMarshal`. we have:

```golang
func nonMontgomeryMarshal(xVal, yVal *fp.Element, output *C.char, outputOffset int) error {
	// Convert g1.X and g1.Y to big.Int using the BigInt method
	var x big.Int
	xVal.BigInt(&x)
	xBytes := x.Bytes()
	xLen := len(xBytes)

	if xLen > 0 {
		// Copy x to output at offset (64 - xLen)
		C.memcpy(unsafe.Pointer(uintptr(unsafe.Pointer(output))+uintptr(outputOffset+64-xLen)), unsafe.Pointer(&xBytes[0]), C.size_t(xLen))
	}

	var y big.Int
	yVal.BigInt(&y)
	yBytes := y.Bytes()
	yLen := len(yBytes)

	if yLen > 0 {
		// Copy y to output at offset (128 - yLen)
		C.memcpy(unsafe.Pointer(uintptr(unsafe.Pointer(output))+uintptr(outputOffset+128-yLen)), unsafe.Pointer(&yBytes[0]), C.size_t(yLen))
	}
	return nil
}
```

If xVal or yVal is zero, then len(xBytes) or len(yBytes) will be zero, which means that the buffer is not written to. This is only safe if the buffer is zero initialized.

kevaundray:
Maybe the java code should zero out the output buffer/ensure it is zeroed out before passing it to the golang code

garyschulte:
> Maybe the java code should zero out the output buffer/ensure it is zeroed out before passing it to the golang code

In AbstractBLS12PrecompiledContract, Besu sends a newly initialized byte array for the output: https://github.com/hyperledger/besu/blob/main/evm/src/main/java/org/hyperledger/besu/evm/precompile/AbstractBLS12PrecompiledContract.java#L111

The default value for a byte is 0 in java, so the byte array is always zero initialized.  I definitely see how this would be a problem if a non-zero initialized array were passed.  Besu's usage is safe, but it is worth noting the zero-initialized requirement.

IMO your added comments are a sufficient call-out for safety here.

garyschulte (approved):

PR #268, opened 2025-03-19
chore: Outline C.memcpy method

kevaundray:
This essentially outlines the one liner method into a CopyBytes and PtrOffset method, since it was not clear at first, what was being memcopied


garyschulte:
LGTM, benchmarks unchanged or better
image

pushed a rebase on 262 for CI reasons

macfarla (approved):

PR #269, opened 2025-03-19
chore: Only use C.int on public cgo methods

kevaundray:
This is a bit of an opinionated PR; where it only uses C.int as an error type in the public methods and all other internal methods use the standard error type


kevaundray:
If we go with #266 this becomes simpler

PR #277, opened 2025-07-03
Optimized ECRECOVER JNA/JNI entrypoints

garyschulte:
This PR adds purpose-specific entrypoints for ECRECOVER for secp256k1 and secp256r1 curves.

The k1 entrypoint is largely just limiting the number of times that besu needs to call across the java/native boundary and results in ~30% performance improvement.  This PR adds a jni source folder for the precompile-specific entrypoints, similar to how gnark and constantine subprojects provide precompile-specific implementations.

The r1 entrypoint uses boringssl to implement an ecrecover entrypoint rather than openssl.  Previously boringssl was used only for p256_verify, this PR consolidates the two precompile entrypoints into a single 'boringssl' library.

Initial integration into besu shows significant improvement in ecrecover on both r1 and k1

```
                               |  Actual cost | Derived Cost |  Iteration time |      Throughput
ecrecover                      |    3,000 gas |    3,186 gas |     31,859.8 ns |      94.16 MGps
ecrecover_perf                 |    3,000 gas |    2,446 gas |     24,464.4 ns |     122.63 MGps

ecrecover_r1 (openssl)         |    3,000 gas |  378,367 gas |  3,783,672.6 ns |       0.79 MGps
ecrecover_r1 (bouncycastle)    |    3,000 gas |  234,788 gas |  2,347,878.9 ns |       1.28 MGps
ecrecover_r1_perf (boringssl)  |    3,000 gas |    5,993 gas |     59,926.4 ns |      50.06 MGps

```

note: this PR has changed CI build machine images to 22.04 as `choosenim` has stopped working on 20.04 images


gradle.properties
```diff
@@ -1 +1 @@
-version=1.3.2-SNAPSHOT
+version=1.3.2-ECRECOVER
```
garyschulte, line 1: delete me

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,159 @@
...
+#include <openssl/ecdsa.h>
+#include <openssl/evp.h>
+#include <openssl/mem.h>
+#include <string.h>
+#include <stdio.h>
+
+#define P256_KEY_LEN 65
+#define P256_COORD_LEN 32
+
+int ecrecover_r1(
```
garyschulte, line 14: This function should be heavily scrutinized. boringssl does not offer public key recovery from signature.  Since we are implementing recovery using boringssl rather than just calling a recovery method, we should scrutinize correctness, edge cases, and performance compared to the openssl implementation.

macfarla (commented):
looks pretty sensible

secp256k1/src/test/java/org/hyperledger/besu/nativelib/secp256k1/ECRecoverPerformanceTest.java
```diff
@@ -0,0 +1,536 @@
...
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Performance comparison between JNA multi-call and JNI single-call ECRECOVER implementations.
+ * This test simulates the ECRECOVER precompile workload using diverse inputs.
+ */
+@RunWith(Parameterized.class)
+public class ECRecoverPerformanceTest {
```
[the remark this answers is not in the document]
lu-pinto: I was going to ask the same thing. Should this not be implemented in the caller of this library instead?
reason: as simple as it may be, it's easy to commit a mistake in coding up a framework like this, so given JMH is already out there it seems reinventing the wheel. Unless you want to JMHify this benchmark

lu-pinto:
Is it worth keeping around `LibSecp256k1.java` ?

secp256k1/src/main/java/org/hyperledger/besu/nativelib/secp256k1/LibSecp256k1JNI.java
```diff
@@ -0,0 +1,98 @@
...
+import org.hyperledger.besu.nativelib.common.BesuNativeLibraryLoader;
+
+public class LibSecp256k1JNI {
+
+  public static final boolean ENABLED;
+
+  static {
+    boolean enabled;
+    try {
+      BesuNativeLibraryLoader.registerJNA(LibSecp256k1JNI.class, "secp256k1_ecrecover");
```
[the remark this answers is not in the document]
garyschulte: it is registerJNA because it is using JNA for library loading and "finding" of the lib in the jar.  We have been using a hybrid JNA/JNI approach, where we get the benefits of jna loading simplicity, but still use the lighter weight JNI calls.

garyschulte:
> Is it worth keeping around `LibSecp256k1.java` ?

I think yes.  We are still using it for block signing and signature verification IIRC.  

If we find that we can benefit from additional performance benefit there we can add that to the jni native bindings.  I might be in the minority on this, but I think having multiple implementations available in besu-native is a good thing.  We don't want too many, but having more than one is good for comparison for performance and correctness at least.

boringssl/src/test/resources/org/hyperledger/besu/nativelib/boringssl/secp256r1ecrecover.csv
```diff
@@ -0,0 +1,49 @@
+input,result,status,error, notes
+9b2db89cb0e8fa3cc7608b4d6cc1dec0114e0b9ff4080bea12b134f489ab2bbc0000000000000000000000000000000000000000000000000000000000000000976d3a4e9d23326dc0baa9fa560b7c4e53f42864f508483a6473b6a11079b2db1b766e9ceb71ba6c01dcd46e0af462cd4cfa652ae5017d4555b8eeefe36e1932,04e266ddfdc12668db30d4ca3e8f7749432c416044f2d2b8c10bf3d4012aeffa8abfa86404a2e9ffe67d47c587ef7a97a7f456b863b4d02cfc6928973ab5b1cb39,0,,pre-existing case from openssl
```
garyschulte, line 2: these input lines are misleading.  we should expect recid to be 27 or 28, and we should subtract 27 (V_BASE) to get 0 or 1 for recid.  these inputs are not wrong for the tests, but because we are encoding recovery id directly as 00..00 or 00..01, these are not realistic precompile inputs.  

we should change the test parser to expect (recovery_id + V_BASE), and do the subtraction before calling the method (as we do in besu).  that way these inputs should be portable into besu r1 tests.

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,159 @@
...
+    int recovery_id,
+    unsigned char output_buffer[65]) {
+
+    if (message_hash == NULL || signature == NULL || output_buffer == NULL) {
+        return 1;
+    }
+
+    if (recovery_id < 0 || recovery_id > 3) {
+        return 1;
+    }
```
garyschulte, line 26: this should be restricted to 0,1.  ecrecover should return uncompressed points, and (2,3) imply compressed points

boringssl/src/test/java/org/hyperledger/besu/nativelib/boringssl/SecP256R1EcrecoverParameterizedTest.java
```diff
@@ -0,0 +1,124 @@
...
+    public String error;
+
+    @Parameterized.Parameter(4)
+    public String notes;
+
+    @Parameterized.Parameters
+    public static Iterable<String[]> parameters() throws IOException {
+        return CharStreams.readLines(
+                new InputStreamReader(
+                        SecP256R1EcrecoverParameterizedTest.class.getResourceAsStream("secp256r1ecrecover.csv"), UTF_8))
```
siladu, line 55: How did you generate the expected outputs?

boringssl/boringssl_jni/ecrecover.h
```diff
@@ -0,0 +1,26 @@
...
+
+// Recovers a P-256 public key from a signature and message hash.
+//
+// All inputs are raw big-endian byte arrays.
+//
+// Returns:
+//   0 - success, public key written to output_buffer
+//   1 - invalid input parameters or signature verification failed  
+//   2 - system error (memory allocation, curve operations failed)
+int ecrecover_r1(
```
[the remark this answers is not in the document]
lu-pinto: this is how the JNI header would look like:
```
JNIEXPORT jint JNICALL Java_EcrecoverJNI_ecrecover_1r1(JNIEnv *, jobject, jbyteArray, jbyteArray, jint, jbyteArray);
```
I personally think that hiding this translation with JNA is not great because I cannot review what I cannot see :)

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,160 @@
...
+    if (message_hash == NULL || signature == NULL || output_buffer == NULL) {
+        return 1;
+    }
+
+    // restrict recovery_id to uncompressed point types
+    if (recovery_id < 0 || recovery_id > 1) {
+        return 1;
+    }
+
+    BN_CTX *ctx = BN_CTX_new();
```
lu-pinto, line 29: I wonder if you should not use `BN_CTX_start` and `BN_CTX_end` instead of creating a new context every time. Possiby you could do `BN_CT_new()` through and `init` native function and then create BIGNUMs out of the `ctx` using `BIGNUM *BN_CTX_get(BN_CTX *ctx);`
I believe if you do this you wouldn't have to free every single individual BIGNUM as they would be freed along with the context. Check out https://docs.openssl.org/1.1.1/man3/BN_CTX_start/#name

Did you look into this alternative?

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,160 @@
...
+    if (!ctx) {
+        return 2;
+    }
+
+    BIGNUM *r = BN_bin2bn(signature, P256_COORD_LEN, NULL);
+    BIGNUM *s = BN_bin2bn(signature + P256_COORD_LEN, P256_COORD_LEN, NULL);
+    BIGNUM *e = BN_bin2bn(message_hash, 32, NULL);
```
lu-pinto, line 36: You probably should get a BIGNUM from the `ctx` and then pass it here instead of `NULL`?

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,160 @@
...
+    int recovery_id,
+    unsigned char output_buffer[65]) {
+
+    if (message_hash == NULL || signature == NULL || output_buffer == NULL) {
+        return 1;
+    }
+
+    // restrict recovery_id to uncompressed point types
+    if (recovery_id < 0 || recovery_id > 1) {
+        return 1;
```
lu-pinto, line 26: `p256_verify.c` has some error printouts for these cases:
```
#define RETURN_INVALID(msg) do { snprintf(error_message_buf, error_message_buf_len, "%s", msg); return 1; } while (0)
#define RETURN_ERROR(msg) do { snprintf(error_message_buf, error_message_buf_len, "%s", msg); return 2; } while (0)
```
should we not do the same for debugging?

boringssl/boringssl_jni/ecrecover.h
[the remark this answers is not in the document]
garyschulte: IMO if we want to change the library loading method to use the super long package space names and JNIEXPORT it should be outside the scope of this PR since it would affect the common library loading code, gnark, constantine, and existing boringssl endpoints.

IMO the shorter name is more readable, not less.  It is not too huge a change to use the gnarly method names instead though.  If we decide to use the gnarly method names,  I'd prefer to make this change comprehensively, rather than in this PR.

secp256k1/secp256k1_jni/secp256k1_ecrecover.c
```diff
@@ -0,0 +1,77 @@
+#include "secp256k1_ecrecover.h"
+#include <secp256k1.h>
+#include <secp256k1_recovery.h>
+#include <string.h>
+
+// Create a shared context for secp256k1 operations
+static secp256k1_context* ctx = NULL;
```
lu-pinto, line 7: is there a concern about multiple threads sharing this context since parallel txs is now available?

garyschulte: Fantastic point.  I hadn't been considering the thread safety of the context for parallel execution.  If you look at the k1 single entrypoint version I am using a single static context.  I am looking into the thread safety of each (secp256k1 and boringssl) and will ensure they are either threadsafe and shared or not threadsafe and created each invocation.  

FWIW, this could be a concern for the existing besu k1 ecrecover implementation already since it uses a single static context

garyschulte: bitcoin-core secp256k1 marks the context as threadsafe, but highlights the ref should be static:
https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L41-L50

I think the current usage is correct and performant for secp256k1.  

BoringSSL bignumber context MIGHT be reusable, but the docs indicate the intent is to create and free it on every invocation:
https://boringssl.googlesource.com/boringssl/+/refs/heads/main/include/openssl/bn.h#251

The libs use their context differently.  Bitcoin-core is expecting it to be long lived since it maintains random seed data.  Boringssl bignumber context is intended to be more ephemeral.  

TL/DR, how we are treating context is threadsafe and idiomatic in both cases.  Thx for raising thread safety issues

boringssl/boringssl_jni/ecrecover.c
garyschulte: Good idea.  I added get from ctx and the requisite start and end/free cleanup.  

Also, bn_ctx related - in p256_verify we are making two different calls that are implicitly creating a BN_CTX: EC_POINT_is_on_curve() and EC_POINT_oct2point().  I refactored a bit to allocate one once and use it for both calls.

I am stopping short of trying to share BN_CTX across invocations, just to keep things simple and clearly threadsafe.

boringssl/boringssl_jni/ecrecover.c
garyschulte: Yup, added similar error messaging and macros for readability.

boringssl/boringssl_jni/ecrecover.c
garyschulte: I am going to keep-it-simple for this PR and we can revisit trying to reuse bn_ctx in subsequent PRs if we need to squeeze the extra juice.

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,160 @@
...
+
+    EC_GROUP *group = EC_GROUP_new_by_curve_name(NID_X9_62_prime256v1);
+    if (!group) {
+        BN_free(r);
+        BN_free(s);
+        BN_free(e);
+        BN_CTX_free(ctx);
+        return 2;
+    }
+    const BIGNUM *order = EC_GROUP_get0_order(group);
```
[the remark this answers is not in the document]
garyschulte: The bignumber is owned/managed by EC_GROUP.  It will be freed by `EC_GROUP_free(group);`

boringssl/boringssl_jni/ecrecover.h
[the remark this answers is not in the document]
lu-pinto: I personally think there are other benefits like input validation that once you translate to pure native types you loose the context and ability to do so in some cases (length checking)

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,160 @@
...
+#include <stdio.h>
+
+#define P256_KEY_LEN 65
+#define P256_COORD_LEN 32
+
+int ecrecover_r1(
+    const unsigned char message_hash[32],
+    const unsigned char signature[64],
+    int recovery_id,
+    unsigned char output_buffer[65]) {
```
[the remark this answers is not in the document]
lu-pinto: well currently there's validation happening here, so there's an argument that the validation is then split between Java and this library, which is not great. I would just move all validation onto here TBH since it's a separate repo. Not a blocker since we are already passing the correct values in Java so maybe address it in another PR.

boringssl/hs_err_pid22360.log
siladu, line 1: committed by mistake?

boringssl/boringssl_jni/ecrecover.c
```diff
@@ -0,0 +1,144 @@
...
+    if (!R) {
+        RETURN_ERROR("EC_POINT allocation failed");
+    }
+    if (!EC_POINT_set_compressed_coordinates_GFp(group, R, r, recovery_id & 1, ctx)) {
+        RETURN_INVALID("failed to recover point R from signature");
+    }
+
+    u1 = BN_new();
+    u2 = BN_new();
+    e_r_inv = BN_new();
```
lu-pinto, line 98: Why not `BN_CTX_get()` on all of these too?

lu-pinto (commented):
Last few comments but looks solid from my side. I even checked that the algorithm match the maths in `ecrecover.c` 👍

boringssl/boringssl_jni/p256_verify.c
```diff
@@ -54,16 +73,16 @@ int p256_verify(
...
+  if (!EC_POINT_is_on_curve(group, point, ctx)) {
     RETURN_INVALID("public key not on curve");
   }
 
   if (EC_KEY_set_public_key(ec_key, point) != 1) {
-    RETURN_ERROR("failed to assign public key to EC_KEY");
+    RETURN_ERROR("failed to assign public key");
   }
 
   r = BN_bin2bn((const uint8_t *)signature_r, P256_COORD_LEN, NULL);
```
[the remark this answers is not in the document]
garyschulte: r and s are owned by ECDSA_SIG, if we use the context, there will be two objects trying to manage/free them.

siladu:
@garyschulte Kind of a naming nitpick. I think the r1 code could be more obviously named r1 to distinguish from k1...

Existing K1: LibSecp256k1.<native>secp256k1_ecdsa_recover within `secp256k1` module
Existing R1: LibSECP256R1.keyRecovery -> BesuNativeEC.<native>p256_key_recovery all within `secp256r1` module

Perf K1: LibSecp256k1JNI.ecrecover -> LibSecp256k1JNI.<native>secp256k1_ecrecover_jni within `secp256k1` module
Perf R1: BoringSSLPrecompiles.ecrecover -> BoringSSLPrecompiles.<native>ecrecover_r1 with `boringssl` module

So for K1 and existing R1, within Besu it is obvious whether we are calling K1 or R1 entry points, and on the besu-native side the module makes it clear (although `BesuNativeEC.p256_key_recovery` isn't clearly r1 without looking at the module).

However for Perf R1, the only mention of r1 is the inner native function name. Especially thinking of the Besu side, BoringSSLPrecompiles might imply R1, but could be more obvious for the uninitiated IMO.

Furthermore, the ecrecover_r1 native code is contained with a file named `ecrecover.c` within `boringssl` module, so not immediately obvious if that is k1 or r1 unless you understand boringssl implies r1.

boringssl/boringssl_jni/ecrecover.c
lu-pinto: Though I feel this might fall into a review blackhole. I opened an issue to track it: https://github.com/hyperledger/besu-native/issues/280

boringssl/boringssl_jni/p256_verify.c
```diff
@@ -74,21 +93,34 @@ int p256_verify(
...
+    ret = 0;
   } else if (verify_status == 0) {
-    RETURN_INVALID("signature verification failed");
+    snprintf(error_message_buf, error_message_buf_len, "%s", "signature verification failed");
+    ret = 1;
   } else {
-    RETURN_ERROR("internal error during signature verification");
+    snprintf(error_message_buf, error_message_buf_len, "%s", "internal error during signature verification");
+    ret = 2;
   }
```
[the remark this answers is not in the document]
garyschulte: The existing flow biases toward expecting success.  There are fewer comparisons in a success case, next fewest comparisons when verification fails, and most in an exceptional case.  But of course any perf difference is so minor.

IMO the way it is presently (with the duplicate/explicit set of ret to 0) is more readable 🤷

lu-pinto (approved):

PR #290, opened 2025-09-03
Rename folder with native binaries from 'lib' to 'lib-native'

### LINKED ISSUE #274: Split package issue when using multiple besu-native Jars with the Java Module System
We noticed a "regression" in `1.2.x` and `1.3.x` when patching several besu-native Jars to add a `module-info.class` file to use them on the `--module-path`. It was introduced with #245. We get the following error:

```
Error occurred during initialization of boot layer
java.lang.LayerInstantiationException: Package lib.aarch64 in both module org.hyperledger.besu.nativelib.secp256k1 and module org.hyperledger.besu.nativelib.arithmetic
```

[Gradle build scan with failing test runs](https://scans.gradle.com/s/mep6qzj3qg6w4)

The reason is that, in the Module System (--module-path), folders in Jar files are interpreted as **Java packages** even if they do not contain `class` files. Then, the strict package visibility rules of the Module System apply which do not allow the same package in multiple Jars (modules).

In #245, a folder structure was introduced, where the folder no longer contain a `-` (which is forbidden in package names). Now, the folders with the native binaries are interpreted as **package** by Java several jars share the same packages such as `lib/aarch64`.

A good solution would be to have the native Jars in a dedicated package in each library. like `lib/secp256k1/...`. If that is too complicated to do with the implementation, reintroducing `-` symbols in the parent folder – e.g. `native-libs` should also fix the issue I believe.

Reproducer as Gradle build. In an empty folder, add:
- `src/main/java/module-info.java` (example code)
```
module org.example {
    requires org.hyperledger.besu.nativelib.arithmetic;
    requires org.hyperledger.besu.nativelib.secp256k1;
}
```
- `src/main/java/org/example/App.java` (example code)
```
package org.example;
public class App {
    public static void main(String[] args) { }
}
```
- `settings.gradle.kts` (empty)
- `build.gradle.kts`
```
plugins {
    id("application")
    id("org.gradlex.extra-java-module-info") version "1.12" // for Jar patching
}

repositories {
    maven("https://hyperledger.jfrog.io/artifactory/besu-maven")
    mavenCentral()
}

extraJavaModuleInfo {
    // Add empty module-info.class to Jars
    module("org.hyperledger.besu:arithmetic", "org.hyperledger.besu.nativelib.arithmetic") {}
    module("org.hyperledger.besu:secp256k1", "org.hyperledger.besu.nativelib.secp256k1") {}
}

application {
    mainModule = "org.example"
    mainClass = "org.example.App"
}

dependencies {
    implementation("org.hyperledger.besu:secp256k1:1.3.1")
    implementation("org.hyperledger.besu:arithmetic:1.3.1")
}
```

Then execute `gradle run` to get the _package in both modules_ error.

jjohannes:
A folder with a `-` in its name is **not** considered a package by Java and is then allowed to exist in multiple Jars without causing a split-package issue. See description of #274 for a detailed explanation.

Fixes #274

### DIFF
--- a/common/src/main/java/org/hyperledger/besu/nativelib/common/BesuNativeLibraryLoader.java
+++ b/common/src/main/java/org/hyperledger/besu/nativelib/common/BesuNativeLibraryLoader.java
@@ -40,7 +40,10 @@ public class BesuNativeLibraryLoader {
   public static void registerJNA(Class jnaClass, String libraryName) {
 
     try {
-      final Optional<Path> libPath = extract(jnaClass, libraryName);
+      String moduleContextClassName = Thread.currentThread().getStackTrace()[2].getClassName();
+      Class<?> moduleContextClass = Class.forName(moduleContextClassName);
+
+      final Optional<Path> libPath = extract(moduleContextClass, libraryName);
 
       if (libPath.isPresent()) {
         NativeLibrary lib = NativeLibrary.getInstance(libPath.get().toString());
@@ -55,6 +58,8 @@ public class BesuNativeLibraryLoader {
                 "Couldn't load native library (%s). It wasn't available at %s or the library path.",
                 libraryName, asLibraryResourcePath(libraryName));
         throw new RuntimeException(exceptionMessage);
+    } catch (ClassNotFoundException e) {
+        throw new RuntimeException(e);
     }
   }
 


jjohannes:
There is another issue when loading on the _module path_. [This code](https://github.com/hyperledger/besu-native/blob/main/secp256r1/src/main/java/org/hyperledger/besu/nativelib/secp256r1/besuNativeEC/BesuNativeEC.java#L30) leads to the attempt to find a resource that is in another Module:

```
BesuNativeLibraryLoader.registerJNA(Library.class, "besu_native_ec_crypto");
```
- `Library.class` is inside `com.sun.jna`
- `besu_native_ec_crypto` is in `org.hyperledger.besu.nativelib.secp256r1`

The loading is later attempted via `classResource.getResourceAsStream(libraryResourcePath)`. That does not work if class and resource are not in the same module.

I am not sure if the code is correct in general. Is it correct that the `besu_native_ec_crypto` implementation is bound to the generic `com.sun.jna.Library`? 
If that is intended, `registerJNA` should probably take two _Class_ arguments:
1. The interface to bind the implementation to
2. A class/interface as a reference to the module the ressource is loaded from via `Class.getResourceAsStream`

Ferparishuertas:
@lu-pinto @macfarla @garyschulte anyways we can bump this PR, and give it  priority?

PR #298, opened 2025-10-16
Static libs and jar artifact for boringssl

garyschulte:
Fixes #295 

This PR adds a new artifact to besu-native builds for p256_verify in boringssl - a static lib, and graal specific classes which enable static compilation. This artifacts target static compilation in order to support besu zkvm guest program development.

This PR builds on top of https://github.com/hyperledger/besu-native/pull/297, and should be reviewed/merged after in order to reduce review burden

Subsequent libraries will be delivered incrementally in a similar fashion.

This PR explicitly does not add tests for the static artifacts, since graal is required. However, tests for the gnark static lib exist in [besu-zkvm](https://github.com/Consensys/besu-zkvm/pull/5)


gradle.properties
```diff
@@ -1 +1 @@
-version=1.4.1-SNAPSHOT
+version=1.4.1-SNASPSHOT
```
garyschulte: fixme

jflo (approved):

